PR #1, opened 2025-06-17
[DPE-7476] Basic charm structure

Zvirovyi:
Works:
* Single unit deployment
* Config change (`profile` only, as dynamic `cluster_name` requires more effort to bring and unknown if needed entirely)
* Unit statuses (including health check)

This PR is reviewable, but lacks:
* Docstrings (see [mattermost](https://chat.canonical.com/canonical/pl/yin7hnedstdht8sifskuaft17o))
* Debug messages
* Documentation (including contributing one with architecture explanation) - **in progress**

@marcoppenheimer


tox.ini
```diff
@@ -0,0 +1,77 @@
...
+    {[testenv]allowlist_externals}
+    find
+commands_pre =
+    poetry install --only lint,format --no-root
+commands =
+    poetry check --lock
+    poetry run codespell {[vars]all_path}
+    poetry run ruff check {[vars]all_path}
+    poetry run ruff format --check --diff {[vars]all_path}
+    find {[vars]all_path} -type f \( -name "*.sh" -o -name "*.bash" \) -exec poetry run shellcheck --color=always \{\} +
```
Batalex, line 49: **praise**: this is nice, we forget too often to check shell scripts

tox.ini
```diff
@@ -0,0 +1,77 @@
...
+    poetry run coverage run --source={[vars]src_path} \
+                 -m pytest \
+                 --tb native \
+                 -v \
+                 -s \
+                 {posargs} \
+                 {[vars]tests_path}/unit
+    poetry run coverage report
+
+[testenv:integration]
```
Batalex: **note**: This is fine right now, but as soon as we get multiple integration test files, we should have multiple tox environments using substitution variables. This will make it so much easier to parallelize the tests on CI

tox.ini
```diff
@@ -0,0 +1,77 @@
...
+    PYTHONPATH
+    CHARM_BUILD_DIR
+    MODEL_SETTINGS
+allowlist_externals =
+    poetry
+
+[testenv:format]
+description = Apply coding style standards to code
+commands_pre =
+    poetry install --only format --no-root
```
Batalex: **nitpick**: I believe you can remove `--no-root` since you have `package-mode=false` in `pyproject.toml`

pyproject.toml
```diff
@@ -0,0 +1,100 @@
...
+
+[tool.poetry.group.integration]
+optional = true
+
+[tool.poetry.group.integration.dependencies]
+pytest = "^8.3.5"
+pytest-asyncio = "*"
+pytest-operator = "^0.42.0"
+# renovate caret doesn't work: https://github.com/renovatebot/renovate/issues/26940
+juju = "<=3.6.1.1"
```
Batalex: **todo**: `pytest-operator` is being deprecated, we should start writing integration tests on a new charm using [jubilant](https://github.com/canonical/jubilant).

To do so, you can start with this steps:
- Add a `tests/integration/conftest.py` file with the following minimal content

```py
from pathlib import Path

import jubilant
import pytest

@pytest.fixture(scope="module")
def juju(request: pytest.FixtureRequest):
    keep_models = bool(request.config.getoption("--keep-models"))

    with jubilant.temp_model(keep=keep_models) as juju:
        juju.wait_timeout = 10 * 60

        yield juju  # run the test

        if request.session.testsfailed:
            log = juju.debug_log(limit=300)
            print(log, end="")

def pytest_addoption(parser):
    parser.addoption(
        "--keep-models",
        action="store_true",
        default=False,
        help="keep temporarily-created models",
    )

@pytest.fixture(scope="module")
def cassandra_charm() -> Path:
    """Path to the packed cassandra charm."""
    if not (path := next(iter(Path.cwd().glob("*.charm")), None)):
        raise FileNotFoundError("Could not find packed cassandra charm.")

    return path
```

- Rewrite `tests/integration/test_charm.py` as follows
```py
import pytest
import yaml

import jubilant

logger = logging.getLogger(__name__)

METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
APP_NAME = METADATA["name"]


def test_deploy(juju: jubilant.Juju, cassandra_charm: Path)->None:
    """Deploy the charm-under-tests together with related charms.
    Assert on the unit status before any relations/configurations take place.
    """
    # Deploy the charm and wait for active/idle status
    juju.deploy(
        cassandra_charm,
        app=APP_NAME,
        config={"profile": "testing"},
    )
    juju.wait(jubilant.all_active)
```

- Add the `-x` pytest option in `tox.ini` to stop the integration tests at the first failure
- Include a build step in the CI to build the charm before running the tests. I suggest having a look at something like https://github.com/canonical/spark-integration-hub-k8s-operator/blob/main/.github/workflows/ci.yaml#L45 followed by https://github.com/canonical/spark-integration-hub-k8s-operator/blob/main/.github/workflows/ci.yaml#L83. That way, you build the charm once, and reuse it on multiple parallel jobs

Batalex (commented):
Will revisit later

src/common/config.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
```
deusebio: This should go to `core`. `common` should be for things that may be common across multiple charms and product, or things that are very very general, with very general signature. For instance, the `cassandra_client` could also be fine in common since it is basically a wrapper of `cassandra` python package, providing context functionalities, which ideally could also be used outside of the Cassadra charm. 

But the information here are definitely `core` information

Just to spell it out, the separation of concerns between common, core, managers, events is:
* common => very common functionalities, that can also be used across multiple products repositories (ideally these functionalities could eventually be moved to a "common" shared library)
* core => domain objects used in the context of charm logic of a specific product (say cassandra charms both VM and K8s)
* managers => businss logic that is charm agnostic. Ideally a manager could also be used outside of charm execution, by defining the domain object and calling the various API. 
* handlers => implementation of charm logic, basically the various hooks that are run in the charms, that of course uses functionalities provided by the managers

Also, given this, there should be a clean dependency relation between these submodules:

```
common < core < managers < handlers
```

src/common/exceptions.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
```
deusebio, line 1: **praise** this is a good class for common

src/common/cassandra_client.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
```
deusebio, line 1: **thought** Although this class looks rather general, I'm not sure why decoupling this from the manager implementation at this point. It is not a strong concern, but I would start simple and bundle this together with the manager. But up to you.

LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
```
welpaolo: I believe we should leave as it is.

metadata.yaml
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: cassandra
+display-name: Cassandra VM
```
welpaolo:
```suggestion
display-name: Apache Cassandra
```

metadata.yaml
```diff
@@ -0,0 +1,13 @@
...
+summary: |
+  Machine charm for Cassandra
```
welpaolo:
```suggestion
  Charmed Apache Cassandra Operator
```

metadata.yaml
```diff
@@ -0,0 +1,13 @@
...
+description: |
```
welpaolo:
```suggestion
description: |
Apache Cassandra is a free and open-source database management system designed to handle large volumes of data across multiple commodity servers. This charm deploys and operate Apache Cassandra on a VM machine environment. 

Apache Cassandra is a free, open source software project by the Apache Software Foundation. 
  Users can find out more at the [Apache Cassandra project page](https://cassandra.apache.org/_/index.html).
```

tox.ini
Zvirovyi: It's not our work :)
`postgresql-operator`, `mysql-operator`, `opensearch-operator` and `charmed-etcd-operator` already had this.

tox.ini
Zvirovyi: Done

tox.ini
Zvirovyi: Done

pyproject.toml
Zvirovyi: Done. CI will be added in the next PR by @BON4.

lib/charms/data_platform_libs/v0/data_models.py
deusebio, line 1: **todo** I believe we have recently released a v1 version that is based on pydantic 2 (see https://github.com/canonical/data-platform-libs/pull/201). Still using pydantic v1 would create a technical debt, so we should change sooner rather than later

metadata.yaml
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 Canonical Ltd.
```
deusebio: **suggestion** the new standard is becoming to bundle all these information of `metadata.yaml`, `config.yaml` and `actions.yaml` into `charmcraft.yaml`. A number of charms in DP still have the "legacy" split into several files, so I don't see very strong drawbacks to keep it this way (hence the suggestion tag) but maybe here we could already start with the new format given we are starting from scratch a new product

src/common/literals.py
deusebio, line 1: two points here:

1. I have not seen any other charm to put a similar files (sometimes called `literals.py` sometimes called `constants.py`) under `common`, and I'm not sure why that should be the case. 
2. A few constants there can probably be placed in the appropriate modules. For instance the one of the paths (e.g. `SNAP_VAR_CURRENT_PATH` through to `MGMT_API_DIR`) probably belongs to the `workload.py` module. Why do they need to have a broader scope?
3. The classes definition should probably be moved to the `core` module (they seem to be domain object), possibly under the `state` submodule

Overall, if any, I would only possibly keep the most high-level variables, e.g. `PEER_PORT` `CLIENT_PORT` and `CLIENT_MGMT_URL` and `SNAP_NAME` (and also the `SNAP_REVISION`) in this file. And move this file probably at root level, like in Kafka or Spark History server.

src/common/statuses.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
```
deusebio: **todo** this module definitely looks like `core`. It seems to be a domain object / definition, and it does pertain to Cassandra specific domains, not very common

src/core/charm.py
deusebio, line 1: **todo** I would just move this class into the root folder. Again, this class does not seem much a domain object. Also, what is the benefit of defining the `CassandraCharmBase` that then gets inherited at root only?

Most importantly, the current implementation breaks the dependency hierarchy 

```
common < core < managers < handlers < root classes
```

This class belongs to `core` but it imports from both managers and handlers.

src/managers/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def reconcile(self, config: CharmConfig):
+        """TODO."""
+        self._render_cassandra_env_config(
+            max_heap_size_mb=1024 if config.profile == "testing" else None,
+        )
+        self._render_cassandra_config(config.cluster_name)
+
+    def _render_cassandra_config(self, cluster_name: str) -> None:
+        config_properties = yaml.safe_load(self.workload.read_file(CAS_CONF_FILE))
```
deusebio: **suggestion** for the files, in Kafka, we use a nice class to bring the various path around, together with the workload, see here: https://github.com/canonical/kafka-operator/blob/main/src/core/workload.py#L18

So basically you can just pass workload, and get both IO operations (e.g. `.read_file`) but also the set of workload files, accessible via `workload.paths.<attribute>`

src/managers/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+        config_properties.update({"cluster_name": cluster_name})
+
+        self.workload.write_file(
+            yaml.dump(config_properties, allow_unicode=True, default_flow_style=False),
+            CAS_CONF_FILE,
+        )
+
+    def _render_cassandra_env_config(self, max_heap_size_mb: int | None) -> None:
+        content = self.workload.read_file(CAS_ENV_CONF_FILE)
```
deusebio: **todo** I'm not sure I'm very keen on the append mode. I would really rather prefer to have an idempotent process, where you don't read from the file system the current file, and append, but you generate it all. Also, expecially when debugging, we often found very useful to have a property representing the file. 

Please have a look at kafka for an example on how to handle configuration files: https://github.com/canonical/kafka-operator/blob/main/src/managers/config.py#L668

src/workload.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            self._cassandra.start(services=[SNAP_SERVICE])
+        except snap.SnapError as e:
+            logger.exception(f"Failed to start cassandra snap: {e}")
+
+    @override
+    def install(self) -> None:
+        """Install the cassandra snap."""
+        logger.debug("Installing & configuring Cassandra snap")
+        snap.install_local("charmed-cassandra_5.0.4_amd64.snap", devmode=True)
```
deusebio: **todo** we need to install from the snapstore and without devmode

src/workload.py
```diff
@@ -0,0 +1,126 @@
...
+        self._cassandra.connect("process-control")
+        self._cassandra.connect("system-observe")
```
deusebio: **todo** I believe we have more connectors here, once we will install from snapstore.

src/workload.py
```diff
@@ -0,0 +1,126 @@
...
+    @override
+    def alive(self) -> bool:
+        try:
+            return bool(self._cassandra.services[SNAP_SERVICE]["active"])
+        except KeyError:
+            return False
+
+    @override
+    def write_file(self, content: str, file: str) -> None:
+        path = Path(file)
```
deusebio: **todo** I would actually start to use the `pathops` library that should make things much much easier when we will be also evolving to K8s. This should also be the general way forward when dealing with filesystem path in charms

You can see the implementation here: https://github.com/canonical/kafka-operator/pull/334/files

tests/unit/test_charm.py
deusebio, line 1: We should definitely tackle this too. Will this be a separate ticket?

deusebio (changes requested):
First pass. The work is going in the right direction but there a few things we need to improve, starting from the structure. In the comments I hope I provide a bit more clarity on how the charm should be structured. But when unsure, please do check the Kafka charm. Whenever possible, I would like to follow the overall patterns and structure there. Eventually this charm will also be owned by the same team, so we should try to get it as close as possible such that the mental burden of switching context is reduced to a minimum.

src/common/cassandra_client.py
Zvirovyi: This client could be used in both cluster management & client resources management. Will this be any good to put into two distinct managers? If yes, it should be here. If no, we can indeed move this to the managers level.

Zvirovyi: It depends on whether it be any good to have separate managers for cluster management & client resources management. If yes, then it should be here. If no, we can indeed move it to the managers level.
Example of cluster management: changing cluster name (should be done both in config file and through cqlsh).
Example of client resources management: creating a keyspace for client relation.

src/common/config.py
Zvirovyi: Agree, done.

lib/charms/data_platform_libs/v0/data_models.py
Zvirovyi: Done

metadata.yaml
Zvirovyi: Done

src/common/statuses.py
Zvirovyi: Done

src/common/literals.py
Zvirovyi: Agree with 2 & 3 (and done that), but disagree with moving this file to the root. Because we will then need either passing it to the managers through initializer parameters (or distinct class with these constants) or importing them from a root (about which I'm strongly against).

charmcraft.yaml
```diff
@@ -0,0 +1,112 @@
...
+    plugin: nil
+    build-packages:
+      - curl
+    override-build: |
+      # Use environment variable instead of `--break-system-packages` to avoid failing on older
+      # versions of pip that do not recognize `--break-system-packages`
+      # `--user` needed (in addition to `--break-system-packages`) for Ubuntu >=24.04
+      PIP_BREAK_SYSTEM_PACKAGES=true python3 -m pip install --user --upgrade pip==24.3.1  # renovate: charmcraft-pip-latest
+
+      # Use uv to install poetry so that a newer version of Python can be installed if needed by poetry
```
welpaolo, line 52: Do we want to use uv ? or just rely on poetry ?

charmcraft.yaml
```diff
@@ -0,0 +1,112 @@
...
+      # Include requirements.txt in *.charm artifact for easier debugging
+      cp requirements.txt "$CRAFT_PART_INSTALL/requirements.txt"
+
+  # "files" part name is arbitrary; use for consistency
+  files:
+    plugin: dump
+    source: .
+    stage:
+      - LICENSE
+      - charmed-cassandra_5.0.4_amd64.snap
```
welpaolo: Why the snap is here?

src/core/charm.py
Zvirovyi: Done

tests/unit/test_charm.py
Zvirovyi: https://warthogs.atlassian.net/browse/DPE-7599

Zvirovyi: Done

welpaolo (commented):
After Enrico's comments resolved we are mostly good to go. I added few comments

charmcraft.yaml
Batalex: My guess is that it was for fast iteration while working both on the snap and the charm. But yeah, we should probably have a FIXME comment here to remove this entry

charmcraft.yaml
Batalex: uv is only used as a pipx+pyenv equivalent here. We have the same setup on kyuubi, though we don't have the focal base issue with the python version being too old for the newest poetry release. I'd say we should keep that for consistency with the other charms

charmcraft.yaml
```diff
@@ -0,0 +1,112 @@
...
+
+      # If Ubuntu version < 24.04, rustup was installed from snap instead of from the Ubuntu
+      # archive—which means the rustup version could be updated at any time. Print rustup version
+      # to build log to make changes to the snap's rustup version easier to track
+      rustup --version
+
+      # rpds-py (Python package) >=0.19.0 requires rustc >=1.76, which is not available in the
+      # Ubuntu 22.04 archive. Install rustc and cargo using rustup instead of the Ubuntu archive
+      rustup set profile minimal
+      rustup default 1.83.0  # renovate: charmcraft-rust-latest
```
Batalex: **todo**: Let's update this to 1.87.0, 1.83.0 is too old to build some of the newest releases of commonly used dependencies in charms

pyproject.toml
```diff
@@ -0,0 +1,100 @@
...
+[tool.poetry.group.lint.dependencies]
+codespell = "^2.4.1"
+
+[tool.poetry.group.unit]
+optional = true
+
+[tool.poetry.group.unit.dependencies]
+coverage = {extras = ["toml"], version = "^7.8.0"}
+pytest = "^8.3.5"
+pytest-asyncio = "*"
```
Batalex: **todo**: We can probably remove this, same for the integration deps. We most likely don't have a need for aync tests now with jubilant

src/common/literals.py
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""TODO."""
+
+from typing import Literal
+
+PEER_RELATION = "cassandra-peers"
+
+SUBSTRATES = Literal["vm", "k8s"]
```
Batalex: **question**: Do we plan on having a k8s charm?

src/core/config.py
```diff
@@ -0,0 +1,33 @@
...
+    cluster_name: str
+
+    @field_validator("profile")
+    @classmethod
+    def profile_values(cls, value: str) -> str:
+        """Check profile config option is one of `testing` or `production`."""
+        if value not in ["testing", "production"]:
+            raise ValueError("profile should be 'testing' or 'production'")
+
+        return value
```
Batalex: **suggestion**: Let's replace the `str` type hint with a `Literal` if we only have two possible values

src/events/cassandra.py
```diff
@@ -0,0 +1,113 @@
...
+        if not self.state.unit.workload_state == UnitWorkloadState.ACTIVE.value:
+            return
+        self.cluster_manager.restart_node()
+
+    def _on_update_status(self, event: UpdateStatusEvent) -> None:
+        self.cluster_manager.update_network_address()
+
+    def _on_collect_unit_status(self, event: CollectStatusEvent) -> None:
+        try:
+            self.charm.config
```
Batalex, line 99: **thought**: I think the charm would raise on init() because of the TypedCharm inheritance, way before we propertly wrap this in a try except to properly set the status

Batalex (commented):
Added a few more thoughts and comments

src/managers/config.py
Zvirovyi: Done

src/workload.py
Zvirovyi: Done

metadata.yaml
Zvirovyi: I removed it after https://github.com/canonical/cassandra-operator/pull/1#discussion_r2157664092 due to error:
```bash
~ charmcraft pack
Bad charmcraft.yaml content:
- extra inputs are not permitted (in field 'display-name')
```

src/workload.py
Zvirovyi: Done. Also added PR to include this info to snap documentation: https://github.com/canonical/charmed-cassandra-snap/pull/11

src/workload.py
Zvirovyi: Done

charmcraft.yaml
Zvirovyi: Removed after https://github.com/canonical/cassandra-operator/pull/1#discussion_r2157683878

metadata.yaml
Zvirovyi: Done

metadata.yaml
Zvirovyi: Done

src/common/literals.py
Zvirovyi: See https://github.com/canonical/cassandra-operator/pull/1#discussion_r2157698271

charmcraft.yaml
Zvirovyi: Done

pyproject.toml
Zvirovyi: Done

src/events/cassandra.py
Zvirovyi: It will raise an error during accessing this property. It can be viewed in the source code:
```python
class TypedCharmBase(CharmBase, Generic[T]):
    """Class to be used for extending config-typed charms."""

    config_type: Type[T]

    @property
    def config(self) -> T:
        """Return a config instance validated and parsed using the provided pydantic class."""
        translated_keys = {k.replace("-", "_"): v for k, v in self.model.config.items()}
        return self.config_type(**translated_keys)
```

src/core/config.py
Zvirovyi: Done

src/common/cassandra_client.py
deusebio: yes, I was thinking that both operations would be done by one manager, that has admin permissions. But again, we can refactor this later, but if it is something to be used within the same charm (although it may pertain to different operations), my feeling is that its place would still be the `manager` folder. `common` is really for all general components that would be used across product in my view

deusebio: Again, this is a non-blocking comment, so feel free to resolve if you prefer to proceed

Zvirovyi: Okay, let's just defer this move until we actually have manager utilizing it. Because now we can't move this client to the manager that doesn't exist :)

src/managers/config.py
Zvirovyi: Let's continue in [Mattermost thread](https://chat.canonical.com/canonical/pl/5hwbzo1zqpfbjj6gs45945d4yy)

src/common/literals.py
deusebio: Yes, I believe you definitely have a point, that root variables should NOT be imported in submodules, but rather root modules should import from the submodules. I believe what I meant was to have it in root and only using it in root (so for instance providing to the classes via constractors arguments), not in other submodules. See if you can do this, otherwise try to move these constants into the dedicated submodules, such that we avoid importing them from root!

src/workload.py
deusebio: Nice! Thanks!

charmcraft.yaml
deusebio: Yes, +1 Alex, makes sense

src/common/literals.py
deusebio: Eventually we will. It is a nice-to-have I'd say, but if we already structure the charm to provide support this (if it is reasonably simply), we will make our life easier in our future

src/events/cassandra.py
deusebio: +1 @Zvirovyi . Yes, it only gets raised as soon as you access the config attribute. I don't know whether raising it earlier would be better, but that's up to the charm authors to decide this

src/managers/cluster.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def start_node(self) -> None:
+        """Start a cluster node."""
+        self.workload.start()
+
+    def restart_node(self) -> None:
+        """Restart a cluster node."""
+        self.workload.restart()
+
+    def update_network_address(self) -> bool:
```
deusebio: **todo** I believe that updating databags is not a manager concern. I'm specifically about line 38:

```
self.state.unit.ip, self.state.unit.hostname = self._network_address()
```

src/managers/config.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(
+        self,
+        workload: WorkloadBase,
+    ):
+        self.workload = workload
+
+    def render_cassandra_config(self, cluster_name: str) -> None:
+        """TODO."""
+        config_properties = yaml.safe_load(self.workload.cassandra_paths.config.read_text())
```
deusebio: **todo** Discussed similarly to the `cassandra-env.sh` file

deusebio (commented):
Getting there. Just a couple of open points but I feel we can wrap up this quite quickly. Ping me on MM to re-review such that we move the basic skeleton quickly

src/common/literals.py
Zvirovyi: Done

src/managers/cluster.py
Zvirovyi: Done

src/managers/config.py
Zvirovyi: Done

src/managers/config.py
Zvirovyi: Done, but waiting for https://github.com/canonical/charmed-cassandra-snap/pull/12 to be merged and published (and for charm's `SNAP_REVISION` to be updated).

README.md
```diff
@@ -1,2 +1,16 @@
 # cassandra-operator
-Charmed Operator for Apache Cassandra
+
+Charmhub package name: operator-template
+More information: https://charmhub.io/cassandra-operator
```
welpaolo: The link is broken I will remove it from the moment.

README.md
```diff
@@ -1,2 +1,16 @@
...
+
+Describe your charm in one or two sentences.
```
welpaolo: Can you provide a very simple description? You can simply copy and paste the one in the `charmcraft.yaml`

src/common/management_client.py
```diff
@@ -0,0 +1,47 @@
...
+        self.base_url = base_url
+
+    def _request(self, method: Literal["GET"], endpoint: str) -> str:
+        response = requests.request(
+            method=method, url=f"{self.base_url}{endpoint}", timeout=_TIMEOUT
+        )
+        response.raise_for_status()
+        return response.content.decode()
+
+    def is_healthy(self) -> bool:
```
welpaolo, line 35: @deusebio  Should we move away from the management-api in favor of `nodetool` calls ?

src/core/config.py
```diff
@@ -0,0 +1,28 @@
...
+from pydantic import field_validator
+
+ConfigProfile = Literal["testing", "production"]
+
+
+class CharmConfig(BaseConfigModel):
+    """Manager for the structured configuration."""
+
+    profile: ConfigProfile
+    cluster_name: str
```
welpaolo, line 19: Any limitation on the cluster name?

src/core/workload.py
```diff
@@ -0,0 +1,145 @@
...
+
+from abc import ABC, abstractmethod
+from typing import Literal
+
+from charmlibs import pathops
+
+Substrate = Literal["vm", "k8s"]
+
+
+class CassandraPaths:
```
welpaolo, line 15: Can you add docstrings for the different paths ?

welpaolo (approved):
Looks fine with me as first skeleton of the charm. 
Docstring can be improved all over the module but this is not blocking for the moment

src/managers/config.py
Zvirovyi: Done, but requires https://github.com/canonical/charmed-cassandra-snap/pull/13 to be merged and published.

src/core/config.py
Zvirovyi: There is check below:
```python
    @field_validator("cluster_name")
    @classmethod
    def cluster_name_values(cls, value: str) -> str:
        """TODO."""
        if len(value) == 0:
            raise ValueError("cluster_name cannot be empty")

        return value
```

README.md
Zvirovyi: Done

README.md
Zvirovyi: Done

LICENSE
Zvirovyi: Done

src/common/management_client.py
deusebio: Yes, we should probably do things without. But we can probably address this as separate ticket to not block further this PR which is blocking some other work. 

@Zvirovyi could you just add a ticket on the backlog about this? Just post the ticket in the comment here and the feel free to resolve.

src/core/workload.py
deusebio: I believe we will address the doc string across the charm in a separate ticket. It was discussed in MM

deusebio (approved):
I believe the state is now good enough to be merged! There are few minor points (doc strings, node tool usage, etc) that we can then address as separate tickets. But this should already unblock a bunch of work that is currently stale

src/core/workload.py
```diff
@@ -0,0 +1,146 @@
...
+        data_dir: pathops.PathProtocol,
+    ) -> None:
+        self._env = env
+        self._config_dir = config_dir
+        self._data_dir = data_dir
+
+    @property
+    def env(self) -> pathops.PathProtocol:
+        """TODO."""
+        return self._env
```
marcoppenheimer: **todo:** Am I correct in thinking that this should now just be a hard-coded value?
Is there any reason for it to be an argument to the class?

marcoppenheimer: Same with `config_dir` and `data_dir`

src/managers/cluster.py
```diff
@@ -0,0 +1,43 @@
...
+        self.workload = workload
+        self.management_client = ManagementClient()
+
+    def start_node(self) -> None:
+        """Start a cluster node."""
+        self.workload.start()
+
+    def restart_node(self) -> None:
+        """Restart a cluster node."""
+        self.workload.restart()
```
marcoppenheimer: **todo:** Let's remove these if possible, they're redundant as we can just call the functions themselves. Better to have 1 place to do a thing, rather than 2.

src/managers/cluster.py
```diff
@@ -0,0 +1,43 @@
...
+
+        Returns:
+            bool: True if the cluster or node is healthy.
+        """
+        return self.management_client.is_healthy()
+
+    def network_address(self) -> tuple[str, str]:
+        """TODO."""
+        hostname = socket.gethostname()
+        return socket.gethostbyname(hostname), hostname
```
marcoppenheimer, line 33: **todo:** This should be part of `UnitContext` and should be removed here.

src/managers/config.py
```diff
@@ -0,0 +1,117 @@
...
+            ),
+            "seed_provider": [
+                {
+                    "class_name": "org.apache.cassandra.locator.SimpleSeedProvider",
+                    "parameters": [{"seeds": ",".join(seeds)}],
+                }
+            ],
+            "storage_compatibility_mode": "CASSANDRA_4",
+            "storage_port": 7000,
+        }
```
marcoppenheimer, line 83: **note:** Not for now, but going forwards, let's see if we can group some of these config options into similar clusters of properties, and then merge them later on.

```python
config = [
    self.foo_config,
    self.bar_config,
    ...
]
```

src/managers/config.py
```diff
@@ -0,0 +1,117 @@
...
+            raise ValueError("cassandra_limit_memory_mb should be at least 1024")
+
+        env = self._map_env(self.workload.cassandra_paths.env.read_text().split("\n"))
+
+        if cassandra_limit_memory_mb:
+            env["MAX_HEAP_SIZE"] = f"{cassandra_limit_memory_mb}M"
+            env["HEAP_NEWSIZE"] = f"{cassandra_limit_memory_mb // 2}M"
+        else:
+            env["MAX_HEAP_SIZE"] = ""
+            env["HEAP_NEWSIZE"] = ""
```
marcoppenheimer: **todo:** Similar to my above comment, could we pull this logic out of here, and add a property that's like:
```python
@property
def heap_config():
    return {MAX_HEAP_SIZE:..., HEAP_NEWSIZE:...}
```

Then here in `render_env`, we just join the output of `render_cassandra_config` and `heap_config`.

src/workload.py
```diff
@@ -0,0 +1,144 @@
...
+from charms.operator_libs_linux.v2 import snap
+from typing_extensions import override
+
+from common.exceptions import ExecError
+from core.workload import CassandraPaths, WorkloadBase
+
+SNAP_CURRENT_PATH = "/snap/charmed-cassandra/current"
+SNAP_VAR_PATH = "/var/snap/charmed-cassandra"
+SNAP_VAR_COMMON_PATH = f"{SNAP_VAR_PATH}/common"
+SNAP_VAR_CURRENT_PATH = f"{SNAP_VAR_PATH}/current"
```
marcoppenheimer: **question:** Do we ever need to know `SNAP_CURRENT_PATH`? Can it be removed?

src/workload.py
```diff
@@ -0,0 +1,144 @@
...
+            return False
+
+    @override
+    def write_file(self, content: str, file: str) -> None:
+        path = self.root / file
+        path.parent.mkdir(exist_ok=True, parents=True)
+        path.write_text(content)
+
+    @override
+    def read_file(self, file: str) -> str:
```
marcoppenheimer, line 76: **nitpick:** Up to you, non-blocking, but could we name these `read` and `write`?

marcoppenheimer (approved):
Approving, as the vast majority of it looks outstanding! Great work!

Left some comments, but they're small and shouldn't need to block, only the `todo`'s need addressing before merging.

Really pleased to see it all laid out like this!!

src/core/workload.py
Zvirovyi: https://warthogs.atlassian.net/browse/DPE-7639

src/common/management_client.py
Zvirovyi: https://warthogs.atlassian.net/browse/DPE-7640

src/managers/cluster.py
Zvirovyi: `UnitContext` can be called for the remote units, and this snippet should be executed locally. For this application we have `UnitContext.ip`.

src/core/workload.py
Zvirovyi: Done, but correct me if I'm wrong in [this commit](https://github.com/canonical/cassandra-operator/pull/1/commits/bae4b4cef10a2e8ac958a304ef7791d7eb03b69a).

src/managers/cluster.py
Zvirovyi: Done

src/workload.py
Zvirovyi: If we will use pathops with all the files then we can get rid of these functions entirely.
@deusebio, what do you think?

src/managers/config.py
Zvirovyi: Done

src/workload.py
Zvirovyi: Done

src/managers/cluster.py
marcoppenheimer: Apologies, you're right! Ignore what I said.

src/core/workload.py
marcoppenheimer: Perfect 👍🏾

src/workload.py
deusebio: I would LOVE to get rid of these :D `pathops` was designed exactly to centralize this, and instead of exposing them via the `Workload` class, abstract them in a class that would work both on K8s and VM. So yes, that would be a very sane refactoring. Feel free to keep it in mind for a next interation

marcoppenheimer: I kinda like it, it's more readable to see `self.workload.write`, but it's not an urgent feeling.

src/managers/config.py
Zvirovyi: I think it will be addressed in authentication PR.

PR #3, opened 2025-06-26
[DPE-7512] Multinode deployment

Zvirovyi:
Also resolves [DPE-7515](https://warthogs.atlassian.net/browse/DPE-7515?atlOrigin=eyJpIjoiNWM0ZTVmYWM0NTkxNGE2YWFjOWRmMzEyM2NkYmYxZTUiLCJwIjoiaiJ9) & [DPE-7601](https://warthogs.atlassian.net/browse/DPE-7601?atlOrigin=eyJpIjoiYTJiMzgyYmRlY2M4NDlhMDk1MGRlZjBjZjU5OGY4ZTAiLCJwIjoiaiJ9) & [DPE-7639](https://warthogs.atlassian.net/browse/DPE-7639?atlOrigin=eyJpIjoiYjgwMmRmNGJkNTE4NDkxM2EzYWRkYmJiMTE2ZDBjZTYiLCJwIjoiaiJ9).

This PR:
1. fixes deprecation warnings from the CQL client;
2. adds type hints to several previously raw unit & cluster context fields;
3. adds new status - `waiting for cluster to start`, when non-leader units are waiting for leader unit to initialize cluster;
4. adds `seeds` field in cluster context in order for leader unit to share its peer address (leader = seed until seed management is established);
5. added read/write test into `integration-charm`;
6. added `integration-multinode` & `integration-scaling` & `integration-config` tests;
7. changes `self.workload.restart()` to `self.workload.start()` in `on.start` event;
8. add `collect_app_status` handler, that sets the application status to blocked on invalid config;
9. removes `collect_status` handlers from `charm.py` due to https://github.com/canonical/cassandra-operator/pull/4#discussion_r2169496514;
10. adds docstrings documentation.

@marcoppenheimer 

[DPE-7515]: https://warthogs.atlassian.net/browse/DPE-7515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-7601]: https://warthogs.atlassian.net/browse/DPE-7601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DPE-7639]: https://warthogs.atlassian.net/browse/DPE-7639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -20,6 +20,8 @@ charm-libs:
     version: "1"
   - lib: data_platform_libs.data_interfaces
     version: "0"
+  - lib: rolling_ops.rollingops
+    version: "0"
 
 config:
   options:
@@ -35,6 +37,8 @@ config:
 peers:
   cassandra-peers:
     interface: cassandra_peers
+  bootstrap:
+    interface: rolling_op
 
 parts:
   # "poetry-deps" part name is a magic constant
--- a/src/common/cassandra_client.py
+++ b/src/common/cassandra_client.py
@@ -9,7 +9,8 @@ from contextlib import contextmanager
 from typing import Generator
 
 from cassandra.auth import PlainTextAuthProvider
-from cassandra.cluster import Cluster, Session
+from cassandra.cluster import EXEC_PROFILE_DEFAULT, Cluster, ExecutionProfile, Session
+from cassandra.policies import DCAwareRoundRobinPolicy, TokenAwarePolicy
 
 logger = logging.getLogger(__name__)
 
@@ -18,6 +19,9 @@ class CassandraClient:
     """TODO."""
 
     def __init__(self, hosts: list[str], user: str | None = None, password: str | None = None):
+        self.execution_profile = ExecutionProfile(
+            load_balancing_policy=TokenAwarePolicy(DCAwareRoundRobinPolicy())
+        )
         self.auth_provider = (
             PlainTextAuthProvider(username=user, password=password)
             if user is not None and password is not None
@@ -28,8 +32,13 @@ class CassandraClient:
         return
 
     @contextmanager
-    def _session(self, keyspace: str | None = None) -> Generator[Session]:
-        cluster = Cluster(contact_points=self.hosts, auth_provider=self.auth_provider)
+    def _session(self, keyspace: str | None = None) -> Generator[Session, None, None]:
+        cluster = Cluster(
+            auth_provider=self.auth_provider,
+            contact_points=self.hosts,
+            protocol_version=5,
+            execution_profiles={EXEC_PROFILE_DEFAULT: self.execution_profile},
+        )
         session = cluster.connect()
         if keyspace:
             session.set_keyspace(keyspace)
--- a/src/common/cassandra_client.py
+++ b/src/common/cassandra_client.py
@@ -16,7 +16,7 @@ logger = logging.getLogger(__name__)
 
 
 class CassandraClient:
-    """TODO."""
+    """Cassandra CQL client."""
 
     def __init__(self, hosts: list[str], user: str | None = None, password: str | None = None):
         self.execution_profile = ExecutionProfile(
--- a/src/core/state.py
+++ b/src/core/state.py
@@ -173,7 +173,7 @@ class ClusterContext(RelationState):
 
     @property
     def is_active(self) -> bool:
-        """Is Cassandra cluster state `ACTIVE`."""
+        """Whether Cassandra cluster state is `ACTIVE`."""
         return self.state == ClusterState.ACTIVE
 
 
--- a/src/core/state.py
+++ b/src/core/state.py
@@ -153,6 +153,15 @@ class ClusterContext(RelationState):
         self.app = component
 
     @property
+    def cluster_name(self) -> str:
+        """Established Cassandra cluster name."""
+        return self.relation_data.get("cluster_name", "")
+
+    @cluster_name.setter
+    def cluster_name(self, value: str) -> None:
+        self._field_setter_wrapper("cluster_name", value)
+
+    @property
     def seeds(self) -> list[str]:
         """List of peer urls of Cassandra seed nodes."""
         seeds = self.relation_data.get("seeds", "")
--- a/src/core/state.py
+++ b/src/core/state.py
@@ -5,7 +5,7 @@
 """TODO."""
 
 import logging
-from enum import Enum
+from enum import StrEnum
 
 from charms.data_platform_libs.v0.data_interfaces import (
     Data,
@@ -22,15 +22,17 @@ CLIENT_PORT = 9042
 logger = logging.getLogger(__name__)
 
 
-class ClusterState(Enum):
+class ClusterState(StrEnum):
     """TODO."""
 
+    UNKNOWN = ""
     ACTIVE = "active"
 
 
-class UnitWorkloadState(Enum):
+class UnitWorkloadState(StrEnum):
     """TODO."""
 
+    UNKNOWN = ""
     STARTING = "starting"
     ACTIVE = "active"
 
@@ -122,11 +124,9 @@ class UnitContext(RelationState):
         return f"{self.ip}:{CLIENT_PORT}"
 
     @property
-    def workload_state(self) -> UnitWorkloadState | None:
+    def workload_state(self) -> UnitWorkloadState:
         """TODO."""
-        if not self.relation or not (value := self.relation_data.get("workload_state", "")):
-            return None
-        return UnitWorkloadState[value.upper()]
+        return self.relation_data.get("workload_state", UnitWorkloadState.UNKNOWN)
 
     @workload_state.setter
     def workload_state(self, value: UnitWorkloadState) -> None:
@@ -148,20 +148,17 @@ class ClusterContext(RelationState):
     @property
     def seeds(self) -> list[str]:
         """TODO."""
-        if not self.relation or not (value := self.relation_data.get("seeds", "")):
-            return []
-        return value.split(",")
+        seeds = self.relation_data.get("seeds", "")
+        return seeds.split(",") if seeds else []
 
     @seeds.setter
     def seeds(self, value: list[str]) -> None:
         self._field_setter_wrapper("seeds", ",".join(value))
 
     @property
-    def state(self) -> ClusterState | None:
+    def state(self) -> ClusterState:
         """The cluster state ('new' or 'existing') of the cassandra cluster."""
-        if not self.relation or not (value := self.relation_data.get("cluster_state", "")):
-            return None
-        return ClusterState[value.upper()]
+        return self.relation_data.get("cluster_state", ClusterState.UNKNOWN)
 
     @state.setter
     def state(self, value: ClusterState) -> None:
--- a/src/core/state.py
+++ b/src/core/state.py
@@ -32,7 +32,7 @@ class ClusterState(StrEnum):
 class UnitWorkloadState(StrEnum):
     """TODO."""
 
-    UNKNOWN = ""
+    INSTALLING = ""
     STARTING = "starting"
     ACTIVE = "active"
 
@@ -126,7 +126,7 @@ class UnitContext(RelationState):
     @property
     def workload_state(self) -> UnitWorkloadState:
         """TODO."""
-        return self.relation_data.get("workload_state", UnitWorkloadState.UNKNOWN)
+        return self.relation_data.get("workload_state", UnitWorkloadState.INSTALLING)
 
     @workload_state.setter
     def workload_state(self, value: UnitWorkloadState) -> None:
@@ -164,6 +164,11 @@ class ClusterContext(RelationState):
     def state(self, value: ClusterState) -> None:
         """TODO."""
         self._field_setter_wrapper("cluster_state", value.value)
+
+    @property
+    def active(self) -> bool:
+        """TODO."""
+        return self.state == ClusterState.ACTIVE
 
 
 class ApplicationState(Object):
--- a/src/core/state.py
+++ b/src/core/state.py
@@ -2,7 +2,7 @@
 # Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""TODO."""
+"""Application state definition."""
 
 import logging
 from enum import StrEnum
@@ -16,30 +16,36 @@ from charms.data_platform_libs.v0.data_interfaces import (
 from ops import Application, CharmBase, Object, Relation, Unit
 
 PEER_RELATION = "cassandra-peers"
-PEER_PORT = 7000
-CLIENT_PORT = 9042
+CASSANDRA_PEER_PORT = 7000
+CASSANDRA_CLIENT_PORT = 9042
 
 logger = logging.getLogger(__name__)
 
 
 class ClusterState(StrEnum):
-    """TODO."""
+    """Current state of the Cassandra cluster."""
 
     UNKNOWN = ""
+    """Cassandra cluster isn't yet initialized by the leader unit."""
     ACTIVE = "active"
+    """Cassandra cluster is initialized by the leader unit and active."""
 
 
 class UnitWorkloadState(StrEnum):
-    """TODO."""
+    """Current state of the Cassandra workload."""
 
     INSTALLING = ""
+    """Cassandra is installing."""
     WAITING_FOR_START = "waiting_for_start"
+    """Subordinate unit is waiting for leader to initialize cluster before it starts workload."""
     STARTING = "starting"
+    """Cassandra is starting."""
     ACTIVE = "active"
+    """Cassandra is active and ready."""
 
 
 class RelationState:
-    """Relation state object."""
+    """Basic class for relation bag mapping classes."""
 
     def __init__(
         self,
@@ -54,9 +60,8 @@ class RelationState:
 
     def _field_setter_wrapper(self, field: str, value: str) -> None:
         if not self.relation:
-            logger.warning(
-                f"Field `{field}` were attempted to\
-                be written on the relation before it exists."
+            logger.error(
+                f"Field `{field}` were attempted to be written on the relation before it exists."
             )
             return
 
@@ -70,7 +75,10 @@ class RelationState:
 
 
 class UnitContext(RelationState):
-    """State/Relation data collection for a unit."""
+    """Unit context of the application state.
+
+    Provides mappings for the unit data bag of peer relation.
+    """
 
     def __init__(
         self,
@@ -88,13 +96,8 @@ class UnitContext(RelationState):
 
     @property
     def unit_name(self) -> str:
-        """The id of the unit from the unit name."""
+        """Unit name."""
         return self.unit.name
-
-    @property
-    def node_name(self) -> str:
-        """The Human-readable name for this cassandra cluster node."""
-        return f"{self.unit.app.name}{self.unit_id}"
 
     @property
     def hostname(self) -> str:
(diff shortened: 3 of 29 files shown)


tests/integration/helpers.py
```diff
@@ -0,0 +1,45 @@
...
+) -> Generator[Session, None, None]:
+    execution_profile = ExecutionProfile(
+        load_balancing_policy=TokenAwarePolicy(DCAwareRoundRobinPolicy()),
+        request_timeout=timeout or 10,
+    )
+    auth_provider = (
+        PlainTextAuthProvider(username=user, password=password)
+        if user is not None and password is not None
+        else None
+    )
```
deusebio, line 32: **question** I don't think we have authentication yet in the charm, don't we? I was looking in the code and I'm not sure how we can enable this. Is it enabled by default? If we have not tested this line, we can probably leave the authentication out from the tests too

src/common/cassandra_client.py
```diff
@@ -18,6 +19,9 @@ class CassandraClient:
     """TODO."""
 
     def __init__(self, hosts: list[str], user: str | None = None, password: str | None = None):
+        self.execution_profile = ExecutionProfile(
+            load_balancing_policy=TokenAwarePolicy(DCAwareRoundRobinPolicy())
```
deusebio, line 23: **question** what would be other option for this? is this something that we may want to configure? can you help me understand what the execution profile does? Thanks!

deusebio (approved):
The code looks already very very good! Great job @Zvirovyi !

Also the tests indeed shows that things are working as intended. Well done!

I only have one questions and a super minor point.

deusebio:
Given I'll be off next week, if you'll do substantial rework as follow up from @welpaolo's review for which another round is necessary, please do put @marcoppenheimer as reviewers. Otherwise if Paolo only raises minor with no substantial rework, feel free to merge with my and Paolo's approval

tests/integration/helpers.py
Zvirovyi: It was just made along with the `CassandraClient` class. Authentication will be covered (and this specific part may change / improve) in [distinct ticket](https://warthogs.atlassian.net/browse/DPE-7641?atlOrigin=eyJpIjoiZGQwZmIwZDQ3MTM3NDdlZjhkM2E4MGMzMTE4NzlmNzMiLCJwIjoiaiJ9).

src/common/cassandra_client.py
Zvirovyi: It is a default value, we just setup it explicitly to avoid any deprecation warnings. Other options are listed [here](https://docs.datastax.com/en/developer/python-driver/3.29/api/cassandra/policies/index.html#cassandra.policies.LoadBalancingPolicy), but I don't think we'll want to change it in the nearest future.

Execution policy is just a class, that holds configs for further connection sessions like balancing policy, request timeout, consistency level and row factory (format of query result data).

src/core/state.py
```diff
@@ -122,20 +122,15 @@ def client_url(self) -> str:
...
-    @property
-    def workload_state(self) -> str:
+    def workload_state(self) -> UnitWorkloadState | None:
         """TODO."""
-        if not self.relation:
-            return ""
-        return self.relation_data.get("workload_state", "")
+        if not self.relation or not (value := self.relation_data.get("workload_state", "")):
+            return None
+        return UnitWorkloadState[value.upper()]
```
marcoppenheimer: **todo:** This feels weird to me.

I think it's OK to type-hint this as `UnitWorkloadState` even if it's not explicitly declared here, but instead grabbed from `workload_state` relation data.
I also don't think it's helpful to mix up `None` and `str` types, let's just add an empty string to the enum, possibly also making it a `StrEnum` for flexibility:

Then, this would look like:

```python

class UnitWorkloadState(StrEnum):
    """TODO."""
	UNKNOWN = ""
    STARTING = "starting"
    ACTIVE = "active"

...

@property
def workload_state(self) -> UnitWorkloadState:
    if not self.relation:
        return ""

    return self.relation_data.get("workload_state", UnitWorkloadState.UNKNOWN)
```

If I'm not mistaken, that will also let us do (I think):

```python
if not self.workload_state:
    foo_bar()
```

src/core/state.py
```diff
@@ -151,14 +146,27 @@ def __init__(
         self.app = component
 
     @property
-    def state(self) -> str:
+    def seeds(self) -> list[str]:
+        """TODO."""
+        if not self.relation or not (value := self.relation_data.get("seeds", "")):
+            return []
+        return value.split(",")
```
marcoppenheimer: **todo:** I think here and elsewhere, we can remove the `if not self.relation or not walrus_operator` syntax, because I believe `self.relation_data` will always be 'something', defaulting to an empty-dict.

It's fine to just do:
```python
@property
def seeds(self) -> list[str]:
    seeds = self.relation_data.get("seeds", "")
    return seeds.split(",") if seeds else []
```

src/core/state.py
```diff
@@ -151,14 +146,27 @@ def __init__(
...
+    def seeds(self, value: list[str]) -> None:
+        self._field_setter_wrapper("seeds", ",".join(value))
+
+    @property
+    def state(self) -> ClusterState | None:
         """The cluster state ('new' or 'existing') of the cassandra cluster."""
-        return self.relation_data.get("cluster_state", "")
+        if not self.relation or not (value := self.relation_data.get("cluster_state", "")):
+            return None
+        return ClusterState[value.upper()]
```
marcoppenheimer: **todo:** Same as earlier similar comment.

src/events/cassandra.py
```diff
@@ -50,48 +50,61 @@ def __init__(
...
 
     def _on_install(self, _: InstallEvent) -> None:
         self.workload.install()
-        self.state.unit.workload_state = UnitWorkloadState.STARTING.value
+        self.state.unit.workload_state = UnitWorkloadState.STARTING
 
     def _on_start(self, event: StartEvent) -> None:
         self._update_network_address()
+
+        if not self.charm.unit.is_leader() and self.state.cluster.state != ClusterState.ACTIVE:
```
marcoppenheimer: **suggestion:** Nitpick, non-blocking so up to you.

Personally I'd wrap this state in to it's own property `self.state.cluster.active`, as it means that the developer working on this event handler file doesn't need to be aware of all the various `ClusterState` values, they just need to do if 'the state of the cluster is active'. It's a minor thing though, so not too important.

src/events/cassandra.py
```diff
@@ -50,48 +50,61 @@ def __init__(
...
 
-        if (
-            not self.charm.unit.is_leader()
-            and self.state.cluster.state != ClusterState.ACTIVE.value
-        ):
-            logger.debug("Deferring on_start for unit due to cluster isn't initialized yet")
-            event.defer()
-            return
+        self.workload.start()
+        self.state.unit.workload_state = UnitWorkloadState.ACTIVE
```
marcoppenheimer: **todo:** We should have some check here to check that the service started successfully, before we declare that the workload is actually 'active'.

I don't know how that would look for Cassandra, but I'm sure there's some sort of 'health check' endpoint we can ping? 'localhost:port/isalive' or something.

src/events/cassandra.py
```diff
@@ -50,48 +50,61 @@ def __init__(
...
+    def _on_config_changed(self, _: ConfigChangedEvent) -> None:
         try:
             self.config_manager.render_env(
                 cassandra_limit_memory_mb=1024 if self.charm.config.profile == "testing" else None
             )
+            self.config_manager.render_cassandra_config(
+                cluster_name=self.charm.config.cluster_name,
+                listen_address=self.state.unit.ip,
+                seeds=self.state.cluster.seeds,
+            )
```
marcoppenheimer: **todo:** I've seen these functions called twice now, I think we should wrap them in to their own method in `config_manager` and just call that here, and on `start`.

Something like `self.config_manager.set_config()`, just to DRY it.

marcoppenheimer: I think we can also avoid passing the args each time, as they'll always be the same. We can instead pass any args needed to the init of the config manager, similar in Kafka we have:
```python
        self.config_manager = ConfigManager(
            state=self.charm.state,
            workload=self.workload,
            config=self.charm.config,
        )
```

src/events/cassandra.py
```diff
@@ -100,17 +113,29 @@ def _on_collect_unit_status(self, event: CollectStatusEvent) -> None:
         except ValidationError:
             event.add_status(Status.INVALID_CONFIG.value)
 
-        if self.state.unit.workload_state == "":
+        if self.state.unit.workload_state is None:
             event.add_status(Status.INSTALLING.value)
```
marcoppenheimer: **question:** Am I correct in assuming that 'if we don't have a workload_state, it must be starting up', so we just say it's installing?

If so, could the `UnitWorkloadState.UNKNOWN` that I showed earlier just be `UnitWorkloadState.INSTALLING` anyway? Do we need the empty string there?

tests/integration/test_scaling.py
marcoppenheimer, line 1: **praise:** Good tests 👍🏾

marcoppenheimer (commented):
Great start! Left some fixes to resolve, but the core is excellent IMO!

src/events/cassandra.py
Zvirovyi: About passing state (e.g. databags) - I disagree. It contradicts our previous discussions with @deusebio:
1. https://github.com/canonical/cassandra-operator/pull/1#discussion_r2156322916
    > managers => business logic that is charm agnostic. Ideally a manager could also be used outside of charm execution, by defining the domain object and calling the various API.

2. https://github.com/canonical/cassandra-operator/pull/1#discussion_r2163706540
    > I believe that updating databags is not a manager concern
    
   (I think that this also applies to reading databags or any other interaction with them)

I fully support position on not exposing state to managers, it's a major thing that separates event handler and managers (fully functional entities). This also makes API more convenient and understandable, so you don't have any hidden data passing.

On the other hand, I also have concern about managers with "dynamic" config. For example, continuing previous discussion about [cassandra client vs cassandra manager](https://github.com/canonical/cassandra-operator/pull/1#discussion_r2156327829), CQL credentials can be changed (they will) and if we have a distinct cassandra manager, how should we update credentials passed to manager? I think the best workaround in this situation will be creating a manager through `@property` in event handler every time with current credentials.

And about single method in config manager to update both configs - considering everything I wrote above, do we need a copy-paste method with all the arguments? It will be 2 definitions of every config instead of 1.

src/core/state.py
Zvirovyi: Done

src/core/state.py
Zvirovyi: Done

src/core/state.py
Zvirovyi: Done

src/events/cassandra.py
Zvirovyi: It will throw an exception if snap cannot start service. Original intent was:
1. `on_start` starts the service and exits
2. service actually becomes active after while and update_status (collect_unit_status) catches this and transits unit to the active / idle state.

Do you propose that we should complete start sequence entirely in `on_start` event? Like deferring until it actually becomes ready?

src/events/cassandra.py
Zvirovyi: Nice catch! Done

src/events/cassandra.py
Zvirovyi: Done

src/events/cassandra.py
marcoppenheimer: Good point, if the snap can't start it'll raise, that's fine.

But, if the snap starts, but is constantly in a restart loop because the service keeps exiting (e.g bad config, other runtime problem etc), then it'll pass, but the service isn't actually 'alive and active', does that make sense?

You raise a good question though, let `update-status` detect the failed service, or `start`. IMO, we should do both. I.e not mark the unit as `UnitWorkloadState.ACTIVE` unless it's actually active, but complete the `on_start` hook like this, THEN emit `update-status` to show the actual status of the service.

src/events/cassandra.py
marcoppenheimer: That's OK, I get the intent.

What we could do, to avoid passing the same args again and again, might be to do:

```python
class ConfigManager:
    def __init__(cluster_name: str, listen_address: IPv4 | IPv6, seeds: str) -> None:
        self._cluster_name = cluster_name
        self._listen_address = listen_address
        self._seeds = seeds

    def render_cassandra_config(self, cluster_name: str | None = None, listen_address: IPv4 | IPv6 | None = None, seeds: str | None = None):
        do_stuff(cluster_name or self._cluster_name)

# then, in charm.py

self.config_manager = ConfigManager(
    cluster_name=self.charm.config.cluster_name,
)

self.config_manager.render_cassandra_config()
```

What do you think?

Zvirovyi: It returns us to the "dynamic config" problem. If any of properties do change, how we should handle this?

src/events/cassandra.py
Zvirovyi: Done

src/events/cassandra.py
Zvirovyi: Another way can be creating `CassandraCharmBase`, that have `config_manager` property with all parameters passed.

Zvirovyi:
@BON4 found a bug with cluster inability to bootstrap several units simultaneously. This leads to 1 unit in `-n3` deployment to be stuck in `Token: (node is not joined to the cluster)` status due to `java.lang.RuntimeException: Bootstrap Token collision between /192.168.2.42:7000 and /192.168.2.41:7000 (token 6896642523066149367` error. I'm working on fixing it with "bootstrap locks".

More info: https://issues.apache.org/jira/browse/CASSANDRA-16364.

UPD: fixed with RollingOpsManager.

src/events/cassandra.py
Zvirovyi: In summary after our meeting: we agreed to pass application state into managers, but we shall not modify it.

deusebio: Just to provide more color, I would be fine to also pass more granular properties (e.g. `listen_address`, `cluster_name`), but I'm afraid that the number of arguments may eventually increase too much, and passing the state can be convenient as we can use one object that is more complex and embed a number of information coming from various databags, config options, etc. 

But  the charm state is to be seen as **a static object**. Information that may be dynamic should be provided as arguments to the function, not coming from the constructor. The pattern that suggests @marcoppenheimer above does this, meaning that if `cluster_name` is something that has just been defined in the hook, the manager call's in the hook should provide that when calling `render_cassandra_config`. 

Ideally every hook execution should be organized in this way:

1. Read the state (this is generally done in the constructor of the charm class, but still it gets executed on every hook)
2. Do business-logic by calling managers functionalities
3. Update databags at the end of the hook

therefore confining side-effects (writing out things) only at the end of the hook. If we operate in this way, the state object should always be static.

src/events/cassandra.py
```diff
@@ -37,82 +39,153 @@ def __init__(
...
+
+    def _on_collect_app_status(self, event: CollectStatusEvent) -> None:
+        try:
+            self.charm.config
+        except ValidationError:
+            event.add_status(Status.INVALID_CONFIG.value)
+
+        event.add_status(Status.ACTIVE.value)
+
+    def bootstrap(self, event: RunWithLock) -> None:
```
deusebio: **suggestion** I wasn't sure where this was used, and then I saw that the influence of the RollingOpsManager is a bit divided between the `events/cassandra` and the main charm, where it is the only method of the charm. I don't have strong concerns with this, but It feels a bit weird. I believe that if the rolling ops is to be used only on this class, we could also just move the `RollingOpsManager` instantiation from `charm.py` to the constructor here, such that we don't have to break the logic a bit here and there. 

it is not a strong point, if you prefer this way, I don't have big concerns.

src/events/cassandra.py
```diff
@@ -37,82 +39,153 @@ def __init__(
...
+        except ValidationError as e:
+            logger.debug(f"Config haven't passed validation: {e}")
+            event.defer()
+            return
+
+        self.workload.start()
+        for _ in Retrying(wait=wait_fixed(10)):
+            if self.cluster_manager.is_healthy:
+                self.state.unit.workload_state = UnitWorkloadState.ACTIVE
+                return
```
deusebio: **todo** What happens if we never get to a healthy cluster? Shouldn't we put a maximum number of retries and fail or raise an exception after? We should also log something such that the user is aware that we are stuck here. 

If it is not super simple to put some logic here, and this requires substantial further work (based on the interplay with rolling ops), I would not block this PR and my suggestion would be to have this in a dedicated ticket, but we should improve this

deusebio (commented):
Really good work @Zvirovyi ! Beside checking the diffs in the PR, I also had a look at the codebase and I'm very much pleased on how the code is shaping up here! Well done!

I only have a small point on the retrying logic that I'd like either to be improved (if it is easy) or to be addressed on a separate ticket. Once the conversation is resolved, I'm happy to approve!

src/events/cassandra.py
Zvirovyi: We cannot do the way you suggested as RollingOpsManager executes function directly from the charm and cannot from the CassandraEvents object, this is why we need this proxy function in `charm.py`. You can check actual code [here](https://github.com/canonical/charm-rolling-ops/blob/3dbcf42036c8d79955fb200594eca6611a397671/lib/charms/rolling_ops/v0/rollingops.py#L414).

src/events/cassandra.py
Zvirovyi: Nice question, I'll put it into Benjamin Allot agenda.

src/events/cassandra.py
marcoppenheimer: Can confirm, it's very annoying but any 'rolling-op' func needs to be in `charm.py`

src/events/cassandra.py
Zvirovyi: @marcoppenheimer
Sorry, I didn't catch your solution at first. I agree that this is a way to go and just implemented this.

> constructor of the charm class, but still it gets executed on every hook

@deusebio 
My bad, I totally forgot that environment recreates on every hook. Thanks for clarification!

src/events/cassandra.py
Zvirovyi: So, I significantly simplified bootstrap logic (it just like restart, but I'm not sure with the proper naming) and moved it to the `charm.py` to avoid proxy call. And also, as multiple event handlers can use this bootstrap method, it will be more convenient to be placed in general place to avoid interaction **between** the event handlers.

src/events/cassandra.py
deusebio: to start simple, can we just put a maximum number of retries with exponential backoff  2,4,8,16,32,64? Then we can resolve the comment

Zvirovyi: Done

Zvirovyi:
I believe everything is done now.

<details>
<summary>Also, I fixed and improved unit tests.</summary>
image
image
image
</details>

deusebio (approved):
Just reduce the waiting to 180 please

Zvirovyi:
UPD: [we agreed](https://chat.canonical.com/canonical/pl/ie41fu9m7jbfdnppesoah6jsso) to keep timeout as it is and move this into distinct ticket.

charmcraft.yaml
```diff
@@ -20,6 +20,8 @@ charm-libs:
     version: "1"
   - lib: data_platform_libs.data_interfaces
     version: "0"
+  - lib: rolling_ops.rollingops
+    version: "0"
```
marcoppenheimer, line 24: **praise:** TIL about defining charm-libs in `charmcraft.yaml`!

marcoppenheimer (approved):
It looks great, I'm feeling REALLY confident in the quality and style of it, it makes me so much more relaxed about taking it over in 3 months, great work, thank you :)

charmcraft.yaml
Zvirovyi: It's indeed very easy to just define it here and do `charmcraft fetch-libs` to get or update them)

In PostgreSQL we encountered many troubles with managing the same version of the postgresql lib through vm & k8s repo. Especially when there are 2 entirely different upgrades of lib in both repos simultaneously (like vm introduced new feature 65 -> 66 version, and on k8s was different PR with another feature, that also upgraded 65 -> 66 version). And I'm sure we will not have such an issue here!)

PR #4, opened 2025-06-26
[DPE-7600] Add unit tests for basic skeleton using `scenario`

BON4:
## Summary

This PR adds comprehensive unit tests to validate the behavior of `CassandraCharm` during lifecycle events such as `start`, `config_changed`, and `collect_unit_status`. These tests improve confidence in the charm’s ability to handle configuration updates, leadership conditions, and cluster health.

## What's included

- ✅ `test_start_maintenance_status_when_starting`: Ensures the unit enters `MaintenanceStatus` when Cassandra is starting.
- ✅ `test_start_sets_active_status_when_healthy`: Confirms that the charm enters `ActiveStatus` when Cassandra becomes healthy.
- ✅ `test_start_only_after_leader_active`: Validates that non-leader units only start after the cluster state is marked `active`.
- ✅ `test_start_not_leader_and_cluster_state_not_active`: Ensures Cassandra does not start if the unit is not the leader and the cluster is not yet active.
- ✅ `test_config_changed_invalid_config`: Verifies that the charm enters `BlockedStatus` on invalid config.
- ✅ `test_config_changed_no_restart`: Ensures the charm avoids restarting if the workload state is still `starting`.
- ✅ `test_collect_unit_status_active_but_not_healthy`: Checks that the unit enters `MaintenanceStatus` if marked active but Cassandra is unhealthy.
- ✅ `test_no_restart_if_not_active`: Ensures a restart is not triggered when workload state is not `active`.

@marcoppenheimer


pyproject.toml
```diff
@@ -34,14 +34,18 @@ optional = true
...
+pytest = "^8.4.1"
+pytest-mock = "^3.14.1"
+ops = {version = "^2.22.0", extras = ["testing"]}
 
 [tool.poetry.group.integration]
 optional = true
 
 [tool.poetry.group.integration.dependencies]
-pytest = "^8.3.5"
+pytest-asyncio = "*"
```
Zvirovyi: We already got rid of `pytest-asyncio` & `pytest-operator` from integration tests in #1. And I don't think we need to reenable `asyncio` entirely.

tests/unit/__init__.py
Zvirovyi, line 1: Can this file be safely removed?

tests/unit/test_charm.py
```diff
@@ -3,6 +3,151 @@
 #
 # Learn more about testing at: https://juju.is/docs/sdk/testing
 
+import ops
+import pytest
+from ops import testing
+from unittest.mock import patch, PropertyMock
+from src.charm import CassandraCharm
+from src.core.state import PEER_RELATION
```
Zvirovyi: Don't import from `src`.

tests/unit/test_charm.py
```diff
@@ -3,6 +3,151 @@
...
+
+def test_start_maintenance_status_when_starting():
+    """Charm enters MaintenanceStatus when Cassandra is not yet healthy."""
+    ctx = testing.Context(CassandraCharm)
+    relation = testing.PeerRelation(id=1, endpoint=PEER_RELATION)
+    state_in = testing.State(leader=True, relations={relation})
+
+    with (
+        patch("managers.config.ConfigManager.render_env"),
+        patch("workload.CassandraWorkload.start"),
```
Zvirovyi: `restart` is actually used here. Until corrected in charm codebase, let's just pull it for now. **This also applies to other tests.**
```suggestion
        patch("workload.CassandraWorkload.restart"),
```

tests/unit/test_charm.py
```diff
@@ -3,6 +3,151 @@
...
+        patch("subprocess.run"),
```
Zvirovyi: I don't think subprocess should be here. Especially if we patching `CassandraWorkload .start` or `CassandraWorkload .restart`. **This also applies to other tests.**

tests/unit/test_charm.py
```diff
@@ -3,6 +3,151 @@
...
+    ):
+        state_out = ctx.run(ctx.on.start(), state_in)
+        assert state_out.unit_status == ops.MaintenanceStatus("waiting for Cassandra to start")
+        assert state_out.get_relation(1).local_unit_data.get("workload_state") == 'active'
```
Zvirovyi: `cluster_state` becomes `active` along with `workload_state` if unit is leader, so we can also check it here.
```suggestion
        assert state_out.get_relation(1).local_unit_data.get("workload_state") == 'active'
        assert state_out.get_relation(1).local_app_data.get("cluster_state") == 'active'
```

tests/unit/test_charm.py
```diff
@@ -3,6 +3,151 @@
...
+    state_in = testing.State(leader=True, relations={relation})
+    with (
+        patch("managers.cluster.ClusterManager.is_healthy", new_callable=PropertyMock) as is_healthy,
+        patch("managers.config.ConfigManager.render_env"),
+        patch("workload.CassandraWorkload.start"),
+        patch("subprocess.run"),
+    ):
+        is_healthy.return_value = False
+        state_out = ctx.run(ctx.on.collect_unit_status(), state_in)
+        assert state_out.unit_status == ops.MaintenanceStatus('waiting for Cassandra to start')
```
Zvirovyi: I think we already test this in `test_start_maintenance_status_when_starting`, and this should be more about `update_status`.

 ## Important!
`collect_unit_status` hook executes here, and this logic also allows for `test_start_sets_active_status_when_healthy` to succeed as there is no `is_healthy` checks in vanilla `on.start` handler. @deusebio should we make tests implying `collect_unit_status` execution, or should we disable / patch it?

tests/unit/test_charm.py
```diff
@@ -3,6 +3,151 @@
...
+    relation = testing.PeerRelation(id=1, endpoint=PEER_RELATION, local_unit_data={"workload_state": "starting"})
+    state_in = testing.State(leader=True, relations={relation})
+    with (
+        patch("managers.config.ConfigManager.render_env"),
+        patch("workload.CassandraWorkload.start"),
+        patch("workload.CassandraWorkload.restart") as restart,
+        patch("subprocess.run"),
+    ):
+        state_out = ctx.run(ctx.on.config_changed(), state_in)
+        restart.assert_not_called()
```
Zvirovyi: Isn't this just a copy of `test_config_changed_no_restart`?

Zvirovyi (changes requested):
I left some initial comments on this PR. Thanks!

tests/unit/__init__.py
BON4: Saw this file is present in a lot of charms tests.

tests/unit/test_charm.py
BON4: Yeah left it by mistake

tests/unit/__init__.py
deusebio: if not needed (meaning unittests runs fine without), let's remove. Less is more! :D

deusebio (commented):
This is good work @BON4 ! I have just maually checked the reports and although we may not aim for 100% coverage, there are a couple of obvious places that I believe we could increase a bit the coverage:

```
src/managers/config.py               33     17      8      0    39%   30-85, 91, 103-110, 114-117, 121-123
src/workload.py                     102     55      6      0    44%   46-50, 55-60, 64-67, 71-73, 77-80, 84, 88, 92-93, 98, 102-109, 113-144
```

I'm saying this because when I look at kafka, the numbers there are higher (see [here](https://github.com/canonical/kafka-operator/actions/runs/15879118971/job/44774360591#step:4:479)), and we can probably take a look at the testing we do there to copy and paste things here too. So they would seem low-hanging fruits.

tests/unit/test_charm.py
deusebio: I believe the tests is fine. One thing I'm realizing is that we have broken down the collect_status_event logic in multiple places, both at `charm.py` but also in the `CassandraEvents`. I'm not sure why, but I would probably keep it in the same file. 

Right now it is a bit misleading (and @Zvirovyi is possibly about this) as in fact we do run twp functions, when there is a CollectStatusEvent for the unit ([here](https://github.com/canonical/cassandra-operator/blob/main/src/charm.py#L47) and [here](https://github.com/canonical/cassandra-operator/blob/main/src/events/cassandra.py#L97) )

To keep things simple, I would probably put all the status logic into the same function. But this can be done in a separate ticket I believe

BON4:
@deusebio 
At the moment, I don't see what usefull tests could be written for config and workload, since the charm is currently quite simple. Most of the features from workload are not used. As we add multinode and TLS features, we'll be able to write more comprehensive tests. For now, I've added several tests for the basic functions.

Now coverage is 80%:
```
Name                              Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------
src/charm.py                         26      0      0      0   100%
src/common/exceptions.py              6      0      0      0   100%
src/common/management_client.py      21      4      0      0    81%   32-33, 43, 47
src/core/config.py                   13      1      2      1    87%   26
src/core/state.py                   105     19     10      3    77%   54-58, 61-64, 84, 89, 94, 117, 122, 127, 133, 183-186, 218-221
src/core/statuses.py                  7      0      0      0   100%
src/core/workload.py                 59     16      0      0    73%   25, 30, 35, 40, 45, 57, 62, 71, 81, 90, 95, 100, 109, 118, 130, 135
src/events/cassandra.py              70      8     12      1    89%   55-56, 64-67, 92, 95
src/managers/cluster.py              13      0      0      0   100%
src/managers/config.py               33      1      8      1    95%   122
src/workload.py                     102     36      6      0    61%   46-50, 55-60, 71-73, 77-80, 84, 88, 92-93, 98, 102-109, 138-144
-----------------------------------------------------------------------------
TOTAL                               455     85     38      6    80%

```

tests/unit/test_config.py
```diff
@@ -0,0 +1,137 @@
...
+    
+    # Verify that workload methods were called
+    mock_workload.cassandra_paths.config.write_text.assert_called_once()
+    
+    # Test env config
+    config_manager.render_env(cassandra_limit_memory_mb=2048)
+    
+    # Verify env methods were called
+    mock_workload.cassandra_paths.env.read_text.assert_called_once()
+    mock_workload.cassandra_paths.env.write_text.assert_called_once()
```
Zvirovyi: This whole test is just to assert 2x`write_text` & 1x`read_text`. I don't think it's effective and can be removed from here.
@deusebio, what do you think?

Zvirovyi (commented):
One more question :)

tests/unit/test_config.py
deusebio: If it does not reduce the coverage, otherwise let's leave it in. Sure, code coverage is not everything and sometimes it is an artefact when you mock everything, but it still has the value to calling the apis, such that when we change apis we are very aware.

deusebio (approved):
Looks good! Thank you for increasing the code coverage. Yes, I would not loose our sleep over this, I just spotted the the numbers were a bit low. Anyhow, the more we stick to the other charm implementations, the more we can probably re-use their tests too, and be more effective. 

When implementing the next unit-tests, just keep an eye on the comparison with other charms for low-handing fruits. That's really my main point, I suppose.

tests/unit/test_config.py
Zvirovyi: @BON4, up to you then

welpaolo (approved):

Zvirovyi (approved):
LGTM, nice work!

BON4:
Just added this patches to some of the tests in `test_charm.py` because after merging main branch tests are failed. 
```python
patch("workload.CassandraWorkload.alive"),
patch("subprocess.run"),
```

Zvirovyi (changes requested):
@BON4, the reason why tests were failing is the `def _on_collect_unit_status(self, event: CollectStatusEvent) -> None:` function, that **automatically executes after every hook**. That function calls `workload.CassandraWorkload.alive` and `managers.cluster.ClusterManager.is_healthy`, that are the original source of this issue - you should patch them and not `subprocess` (you already did former, just add latter).

Zvirovyi (approved):

PR #5, opened 2025-06-27
[DPE-7513] Basic github ci pipeline for charm

BON4:
@marcoppenheimer

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,22 +72,13 @@ jobs:
       path-to-charm-directory: ${{ matrix.path }}
 
   integration-test:
-    strategy:
-      fail-fast: false
-      matrix:
-        juju:
-          # This runs on all runs
-          - agent: 3.6.7 # renovate: juju-agent-pin-minor
-            allure_report: false
-    name: Integration test charm | ${{ matrix.juju.agent }}
+    name: Integration test charm
     needs:
       - lint
       - unit-test
       - build
-    uses: canonical/data-platform-workflows/.github/workflows/integration_test_charm.yaml@v29.0.0
+    uses: ./.github/workflows/integration_test.yaml
     with:
-      juju-agent-version: ${{ matrix.juju.agent }}
       artifact-prefix: ${{ needs.build.outputs.artifact-prefix }}
-      cloud: lxd
     permissions:
-      contents: write
+      contents: write  # Needed for Allure Report
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,18 +8,17 @@ concurrency:
   cancel-in-progress: true
 
 on:
+  workflow_call:
+    outputs:
+      artifact-prefix:
+        description: build_charm.yaml `artifact-prefix` output
+        value: ${{ jobs.build.outputs.artifact-prefix }}
   pull_request:
     paths-ignore:
       - 'docs/**'
       - '**.md'      
   schedule:
     - cron: '53 0 * * *'  # Daily at 00:53 UTC
-  # Triggered on push to default branch by .github/workflows/release.yaml
-  workflow_call:
-    outputs:
-      artifact-prefix:
-        description: build_charm.yaml `artifact-prefix` output
-        value: ${{ jobs.build.outputs.artifact-prefix }}
 
 jobs:
   lint:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,93 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: Tests
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '**.md'      
+  schedule:
+    - cron: '53 0 * * *'  # Daily at 00:53 UTC
+  # Triggered on push to default branch by .github/workflows/release.yaml
+  workflow_call:
+    outputs:
+      artifact-prefix:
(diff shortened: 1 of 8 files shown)


Zvirovyi:
Should we use `spread` for integration tests?

deusebio:
@Zvirovyi why not! I believe it is sensible. @marcoppenheimer any thoughts?

Also in this PR, @BON4 please disable caching for the time being (unless you want to integrate the build in charmcraft cache). Given it is still a simple charm, not using caching may not be that critical for now

To see how to disable cache, see [here](https://github.com/canonical/katib-operators/blob/aa012fe9d30a9d955454bf903ddc55f4a48dc088/.github/workflows/ci.yaml#L100)

deusebio:
@BON4 I have re-assigned the review to Bikalpa. Alex is currently out of office until the second week of July

marcoppenheimer:
Haven't reviewed yet, but on the `spread` discussion, yeah let's use it as it's the way forwards. Useful docs based on the 'old' way using data-platform-workflows can be found here - https://github.com/canonical/data-platform-workflows/blob/main/.github/workflows/integration_test_charm_deprecation_notice.md

.github/workflows/release.yaml
deusebio, line 1: **note** I don't think we have created the charmhub space yet, and we can probably leave without one until we have at least authentication in place. I don't think it will be too critical as there are no downstream components using this charm. We can still keep on testing with locally built charms for a few more days/weeks

But we can keep this here, just such that as soon as we will have charmhub and the token in the repo, this will work

deusebio (approved):
Looks good!

deusebio (changes requested):
Sorry I errounously approve. We should get the CI green and make sure that the integration tests pass. I will let @marcoppenheimer chime in if we want to use spead. I would prefer to use spread such that we are already compliant looking forward. 

However, given we possibly need some further work in this PR, I'll hand it over to @marcoppenheimer given I'll be off in the next few days

theoctober19th (approved):
The changes and the code all seem sensible to me. I'm not too hard-set on the usage of the `spread`, I'll leave this to Marc and other reviewers. Let's the CI green, the rest looks good to me.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,92 @@
...
+        uses: canonical/charming-actions/check-libraries@2.7.0
+        with:
+          credentials: ${{ secrets.CHARMHUB_TOKEN }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+    permissions:
+      pull-requests: write
+
+  build:
+    strategy:
+      fail-fast: true
```
theoctober19th, line 64: I think the reason CI is failing now is because the cache is not disabled here (and thus `charmcraft` tries to build this charm looking at its cache in `charmcraftcache-hub`. add `cache: false` here to disable it for now.

theoctober19th: read more here: https://github.com/canonical/data-platform-workflows/blob/main/.github/workflows/build_charm.md

BON4:
@deusebio @marcoppenheimer I've added spread tests.

spread.yaml
```diff
@@ -0,0 +1,132 @@
+project: mysql-router-k8s-operator
```
Zvirovyi: Old project name.

tests/spread/test_charm.py/task.yaml
```diff
@@ -0,0 +1,7 @@
+summary: test_charm.py
+environment:
+  TEST_MODULE: test_charm.py
+execute: |
+  tox run -e integration -- "tests/integration/$TEST_MODULE" --model testing --alluredir="$SPREAD_TASK/allure-results"
```
Zvirovyi:
```suggestion
  tox run -e integration-charm -- --model testing --alluredir="$SPREAD_TASK/allure-results"
```

BON4: Thx!

BON4:
@marcoppenheimer @deusebio @theoctober19th  
What should the `SPREAD_PASSWORD` for GitHub Actions be?  
And who is allowed to set it?

marcoppenheimer:
@BON4 - we can leave it empty, no need to have it set. Example from `mysql` - https://github.com/canonical/mysql-router-k8s-operator/blob/94192d958f32ce35e0dc7737e068eee6c554fab9/spread.yaml#L82

tests/spread/test_charm.py/task.yaml
Zvirovyi: @BON4 `integration-charm` doesn't require you to specify test file as `TEST_MODULE` already included [here](https://github.com/canonical/cassandra-operator/blob/08d218b2c7305e8ad431a50d6e7241f9101b0233/tox.ini#L70).
```shell
$ tox -e integration-charm
```

BON4:
@marcoppenheimer CI is green and tests are passed

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,84 @@
...
+    paths-ignore:
+      - 'docs/**'
+      - '**.md'
+  schedule:
+    - cron: '53 0 * * *'  # Daily at 00:53 UTC
+
+jobs:
+  lint:
+    name: Lint
+    uses: canonical/data-platform-workflows/.github/workflows/lint.yaml@v29.0.0
```
marcoppenheimer, line 26: **note:** It's not part of the default lint check, but I think we need to add a static type checker here. Maybe adding `pyright` to `tox -e lint`. This can be a separate ticket, just noting here.

marcoppenheimer (approved):
Looks good 👍🏾 

Have added `CHARMHUB_TOKEN`, `JIRA_API_TOKEN` and `JIRA_USER_EMAIL` as GH secrets, please ping me if any CI fails due to missing secrets and I can add them.

PR #8, opened 2025-07-03
[DPE-7641] Authentication

Zvirovyi:
# Overview

This PR:
1. Splits config definition in corresponding manager (https://github.com/canonical/cassandra-operator/pull/1#discussion_r2167476495).
2. Extends start logic to configure password for system `cassandra` user. Showcase diagram of changes is presented below.
3. Migrates Cassandra client to new database manager (https://github.com/canonical/cassandra-operator/pull/1#discussion_r2156327829).
4. Improved config changed handler to not execute rolling restart if config file isn't actually changed.

# Diagram of startup logic changes

Diagram describes `on_start` event handling represented from the leader perspective as there's no changes for the subordinate units.

image


src/events/cassandra.py
```diff
@@ -79,20 +86,57 @@ def _on_start(self, event: StartEvent) -> None:
...
             listen_address=self.state.unit.ip,
             seeds=self.state.cluster.seeds,
+            authentication=True,
         )
+        self.charm.on[str(self.bootstrap_manager.name)].acquire_lock.emit()
 
+    def _start_password_change(self, event: StartEvent) -> None:
+        self.config_manager.render_cassandra_config(
+            cluster_name=self.charm.config.cluster_name,
+            listen_address="127.0.0.1",
```
welpaolo: Maybe it should be a good idea to have constants.

src/managers/config.py
```diff
@@ -135,3 +106,76 @@ def _env_heap_config(cassandra_limit_memory_mb: int | None) -> dict[str, str]:
...
+            "network_authorizer": "AllowAllNetworkAuthorizer",
+            "num_tokens": 16,
+            "partitioner": "org.apache.cassandra.dht.Murmur3Partitioner",
+            "replica_filtering_protection": {
+                "cached_rows_fail_threshold": 32000,
+                "cached_rows_warn_threshold": 2000,
+            },
+            "role_manager": "CassandraRoleManager",
+            "storage_compatibility_mode": "CASSANDRA_4",
+            "storage_port": 7000,
```
welpaolo, line 283: Is it maybe a good idea to have all those integer values?

welpaolo (commented):
Good starting point. I thought that this PR was also including the handling of system user password with secrets like [here](https://github.com/canonical/postgresql-operator/pull/833). Will this be part of another PR ?

src/managers/config.py
Zvirovyi: ?

src/events/cassandra.py
Zvirovyi: Is a single use of "127.0.0.1" worth putting it in the constant? Or are you talking about something other?

Zvirovyi:
> Good starting point. I thought that this PR was also including the handling of system user password with secrets like [here](https://github.com/canonical/postgresql-operator/pull/833). Will this be part of another PR ?

Did I understood it correctly, that this question is about secret changed event? I thought I'll put it in another PR, but considering that this PR is relatively small, I'll include it here. Thanks!

charmcraft.yaml
```diff
@@ -26,14 +26,18 @@ charm-libs:
 
 config:
   options:
+    cluster_name:
```
welpaolo: Do we really need this option?

src/events/cassandra.py
```diff
@@ -104,33 +116,105 @@ def _on_start(self, event: StartEvent) -> None:
...
+        return self.workload.generate_password()
+
+    def _start_password_change(self, event: StartEvent) -> None:
+        self.config_manager.render_cassandra_config(
+            cluster_name=self.state.cluster.cluster_name,
+            listen_address="127.0.0.1",
+            seeds=["127.0.0.1:7000"],
+            authentication=False,
+        )
+        self.workload.start()
```
deusebio: **question** If I understand the logic correctly, here we are starting the server with no authentication. This seems weird and it could pose a security threat. I understand that we are possibly restricting the listen_address into localhost, but why do we need to do this?

Another possibility could also be to start the server using a temporary username and password that only the units knows, such that it can still use the service in authenticated fashion, and then change the password to use the "official" one. 

anyhow, let's discuss later at daily

src/charm.py
```diff
@@ -41,10 +42,20 @@ def __init__(self, *args):
             cluster_name=self.state.cluster.cluster_name,
             listen_address=self.state.unit.ip,
             seeds=self.state.cluster.seeds,
-            enable_peer_tls=False,
-            enable_client_tls=False,
+            enable_peer_tls=self.state.unit.peer_tls.ready,
+            enable_client_tls=self.state.unit.client_tls.ready,
             keystore_password=self.state.unit.keystore_password,
             truststore_password=self.state.unit.truststore_password,
+            authentication=bool(self.state.cluster.cassandra_password_secret),
```
deusebio: **quesiton** why do we allow this? Maybe I would not expose this, and only allow authentication to be true, just to prevent mistakes. We must ensure that authentication is enabled at all times, if external access is possible.

src/managers/config.py
```diff
@@ -37,116 +38,65 @@ def __init__(
...
-                "algorithm": "SunX509",
-                "store_type": "JKS",
-                "protocol": "TLS",
-            }
-
         self.workload.cassandra_paths.config.write_text(
-            yaml.dump(config, allow_unicode=True, default_flow_style=False)
+            yaml.dump(
+                self._merge_dicts(
+                    [
```
deusebio: **praise** I really really really like how you splitted the config into separate section. It is a lot more maintanable IMHO

deusebio (commented):
Let's just discuss about the authenitcation flow, to be on the same page. I'll continue with the review after the discussion

src/events/cassandra.py
```diff
@@ -104,33 +116,105 @@ def _on_start(self, event: StartEvent) -> None:
...
+        event.defer()
 
+    def _finalize_password_change(self, event: StartEvent) -> None:
+        if not self.cluster_manager.is_healthy:
+            event.defer()
+            return
+        self.database_manager.update_system_user_password(
+            "cassandra", self.state.cluster.cassandra_password_secret
+        )
+        self.cluster_manager.prepare_shutdown()
```
deusebio: **question** why do we drain before reconfiguring the password? The drai operation may be massive?

src/events/cassandra.py
```diff
@@ -104,33 +116,105 @@ def _on_start(self, event: StartEvent) -> None:
...
+        if self.state.unit.workload_state != UnitWorkloadState.ACTIVE:
+            event.defer()
+            return
         try:
+            if self.charm.unit.is_leader() and self.state.cluster.cassandra_password_secret != (
+                password := self._acquire_cassandra_password()
+            ):
+                self.database_manager.update_system_user_password("cassandra", password)
+                self.state.cluster.cassandra_password_secret = password
+                self.cluster_manager.prepare_shutdown()
```
deusebio: **issue** it is strange that we need to drain for every config change/rotation password

src/events/cassandra.py
Zvirovyi: Done. Now authentication is always enabled, and we just use default credentials instead of forceful password hash update.

src/events/cassandra.py
Zvirovyi: I've improved drain usage.

tests/integration/test_authentication.py
```diff
@@ -0,0 +1,100 @@
...
+
+    juju.deploy(
+        cassandra_charm,
+        app=app_name,
+        config={"profile": "testing", "system_users": custom_secret},
+    )
+
+    juju.cli("grant-secret", "custom_secret", app_name)
+
+    juju.wait(jubilant.all_active)
```
deusebio: **question** aren't we asserting anything here? I was kinda expecting to assert a blocked status given that the name of the tests is `juju_deploy_BAD_custom_secret`

tests/integration/helpers.py
```diff
@@ -18,20 +18,21 @@
...
 def connect_cql(
+    juju: jubilant.Juju,
+    app_name: str,
     hosts: list[str],
-    user: str | None = None,
-    password: str | None = None,
     keyspace: str | None = None,
     timeout: float | None = None,
 ) -> Generator[Session, None, None]:
+    secrets = get_secrets_by_label(juju, f"cassandra-peers.{app_name}.app", app_name)
```
deusebio: **niptick** I know it is a matter of taste, but I would rather place the password as argument. I feel right now we are parsing the secret twice, here and in the outside function (where we also test that it is - or it is not - the value `custom_password`)

But this way it should also be easy to try to connect with the "wrong" password, or the wrong username

tests/integration/test_authentication.py
deusebio, line 1: **todo** it would also be nice to do the unhappy tests, where we try to connect with:

1. the wrong password
2. with a non-existing username

and we get some sort of unauthenticated error

tests/integration/helpers.py
```diff
@@ -18,20 +18,21 @@
...
     execution_profile = ExecutionProfile(
         load_balancing_policy=TokenAwarePolicy(DCAwareRoundRobinPolicy()),
         request_timeout=timeout or 10,
     )
-    auth_provider = (
-        PlainTextAuthProvider(username=user, password=password)
-        if user is not None and password is not None
-        else None
+    auth_provider = PlainTextAuthProvider(
+        username="cassandra", password=secrets[0]["cassandra-password"]
```
deusebio: **todo** can we parametrize for the username to also tests a non-existing user?

deusebio (commented):
Looks good, and I believe it should be closed to be mergeable. There are just a few comments and improvements I'd like to get in before final approval. 

As a side note, although it is fine, I find the startup logic a bit cumbersome and it took me a few read and navigation of the codebase before understanding the bootstrap process logically. 
Most of the logic is indeed in the `_on_start` method, but there are a few things that we could do to improve the readability (but I would let @marcoppenheimer and @welpaolo provide guidance whether this is sensible given that they have way more experience writing - and reading - charms):

1. Clearly split the logic between leader and non-leader. For instance, it took me some time before realizing that 
```
 if self.state.unit.workload_state == UnitWorkloadState.CHANGING_PASSWORD:
``` 
was only relevant for leader units. 

2. We heavily rely on event deferrals to co-ordinate within the same units. For instance we first execute `_start_password_change`, and then we defer, and then we execute (in the next event hook) `_finalize_password_change`. I'm wondering if we could handle both calls in the same event hook, without splitting this between two hooks

3. Try to follow a logical order, since right now the `_start_password_change` is placed after `_finalize_password_change`, which is somewhat counter-intuitive (also here if we were to use the same hook to start and rotate the password, we would probably address this too)

4. **niptick** I would put drain operation and restart together, unless we have good reasons not to. Right now, we drain before the rolling ops acquires the lock, and then restart during the [bootstrap call](https://github.com/canonical/cassandra-operator/pull/8/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR110). I know I raised the concern of not draining too often (like password rotation or simple config change that does not require shutting down the server), but I suppose when shutting down the server, the drain is generally always required, isn't it?

tests/integration/test_authentication.py
Zvirovyi: If secret is bad (secret is inaccessible or doesn't contain correct data), we just skip it. I see no point in blocking charm because of it. On my opinion, the only viable thing to add will be a comprehensive log message about it.

If user fixes an accessible secret, we will receive secret changed event and will use it.

Zvirovyi:
@deusebio Done!

charmcraft.yaml
```diff
@@ -28,14 +28,18 @@ charm-libs:
...
```
welpaolo:
```suggestion
    cluster-name:
```

charmcraft.yaml
```diff
@@ -28,14 +28,18 @@ charm-libs:
...
+      # TODO: description
+      type: string
+      default: Test Cluster
     profile:
       # TODO: description
       type: string
       default: production
-    cluster_name:
+    system_users:
```
welpaolo:
```suggestion
    system-users:
```

src/events/cassandra.py
```diff
@@ -105,32 +140,89 @@ def _on_start(self, event: StartEvent) -> None:
...
 
-    def _on_config_changed(self, _: ConfigChangedEvent) -> None:
-        # TODO: add peer relation change hook for subordinates to update leader address too
+    def _acquire_cassandra_password(self) -> str:
+        if self.charm.config.system_users:
+            try:
+                if (
+                    password := self.model.get_secret(id=self.charm.config.system_users)
+                    .get_content(refresh=True)
+                    .get("cassandra-password")
```
welpaolo: if we want to want to be consist with other charms we should have `username:password` something like: `cassandra:mybeautifulpassword`

src/events/cassandra.py
```diff
@@ -144,10 +236,9 @@ def _on_update_status(self, _: UpdateStatusEvent) -> None:
             self.config_manager.render_cassandra_config(
                 listen_address=self.state.unit.ip,
                 seeds=self.state.cluster.seeds,
-                keystore_password=self.state.unit.keystore_password,
-                truststore_password=self.state.unit.truststore_password,
             )
 
+            self.cluster_manager.prepare_shutdown()
```
welpaolo, line 231: Do we need to `drain` here ?

src/managers/config.py
```diff
@@ -193,3 +169,133 @@ def _env_jmx_exporter_config(
             if agent_path
             else "",
         }
+
+    def _cassandra_client_tls_config(
+        self,
+        enabled: bool,
+        keystore_password: str | None,
+        truststore_password: str | None,
+    ) -> dict[str, Any]:
```
welpaolo, line 161: Can you please add docstring?

tests/integration/test_authentication.py
welpaolo: If the secret is not in the correct format we should put the charm in blocked state following DA146.

welpaolo (commented):
Great work! I left some comments to fix the last bits

src/events/cassandra.py
Zvirovyi: Yes. There is a check `if self._update_network_address()` above, that returns positive only if IP address of the unit is changed. And Cassandra need to be restarted if seeds or own address is changed.

src/managers/config.py
Zvirovyi: Isn't function name self-explanatory? Same with `_cassandra_peer_tls_config`, `_cassandra_authentication_config`, `_cassandra_connectivity_config`, `_cassandra_directories_config`, `_cassandra_default_config`.

src/events/cassandra.py
welpaolo: Ok perfect thanks for confirming :)

src/managers/config.py
welpaolo: It is okay for the time being

charmcraft.yaml
Zvirovyi: I've removed this config option and tied it to the Juju application name for now.

src/charm.py
```diff
@@ -45,13 +46,19 @@ def __init__(self, *args):
...
-            enable_client_tls=False,
+            enable_peer_tls=self.state.unit.peer_tls.ready,
+            enable_client_tls=self.state.unit.client_tls.ready,
             keystore_password=self.state.unit.keystore_password,
             truststore_password=self.state.unit.truststore_password,
+            authentication=True,
+        )
+        database_manager = DatabaseManager(
+            hosts=[self.state.unit.ip],
+            user="cassandra",
```
marcoppenheimer: **todo:** Can we move the `cassandra` string to a variable some `src/literals.py` please? I know we don't have that file, but somewhere to keep globals.

And is this supposed to be the 'charm admin' user? If so, could we rename it to `operator` to adhere with the other DP charms?

src/charm.py
```diff
@@ -93,16 +103,22 @@ def bootstrap(self, event: RunWithLock) -> None:
...
+            wait=wait_exponential(), stop=stop_after_delay(1800), reraise=True
+        ):
+            with attempt:
+                if not self.cluster_manager.is_healthy:
+                    raise Exception("bootstrap timeout exceeded")
+
+        if self.state.unit.peer_tls.rotation:
+            self.state.unit.peer_tls.rotation = False
+        if self.state.unit.client_tls.rotation:
+            self.state.unit.client_tls.rotation = False
```
marcoppenheimer, line 119: **todo:**
```suggestion
        self.state.unit.peer_tls.rotation = not self.state.unit.peer_tls.rotation
        self.state.unit.client_tls.rotation = not self.state.unit.client_tls.rotation
```

src/core/config.py
```diff
@@ -7,22 +7,12 @@
...
-from pydantic import field_validator
 
 ConfigProfile = Literal["testing", "production"]
 
 
 class CharmConfig(BaseConfigModel):
     """Structured charm config."""
 
-    cluster_name: str
+    system_users: str
```
marcoppenheimer, line 17: **todo:** Please add a validator here for `system_users`. You can probably copy https://github.com/canonical/kafka-operator/blob/2c6b6b309944d282eb17c25c39afc5c2731fba44/src/core/structured_config.py#L251-L261 exactly

src/core/state.py
```diff
@@ -46,7 +46,7 @@
     "peer-private-key-secret",
 ]
 
-SECRETS_APP = ["internal-ca-secret", "internal-ca-key-secret"]
+SECRETS_APP = ["internal-ca-secret", "internal-ca-key-secret", "cassandra-password"]
```
marcoppenheimer: **todo:** I think this is the `operator-password` no? `cassandra-password` is a bit ambiguous.

marcoppenheimer: Ah, I scrolled down. Let me re-phrase.

**todo:** Can we use the following users (if possible):
- `operator` - the charm-internal 'admin' user
- `replication` - the charm-internal node-node communication user

src/core/statuses.py
```diff
@@ -17,6 +17,7 @@ class Status(Enum):
     STARTING = MaintenanceStatus("waiting for Cassandra to start")
     WAITING_FOR_CLUSTER = WaitingStatus("waiting for cluster to start")
     INVALID_CONFIG = BlockedStatus("invalid config")
+    INVALID_SYSTEM_USERS_SECRET = BlockedStatus("invalid system-users secret")
```
marcoppenheimer, line 20: **todo:** If we have the `system-users` config validator, we shouldn't need this validation or `BlockedStatus`. We should just let the charm go in to `error` state, and force the user to re-set the config and `juju resolve`.

src/events/cassandra.py
```diff
@@ -68,26 +76,16 @@ def _on_install(self, _: InstallEvent) -> None:
...
-            sans_ip=[host_mapping[0]],
-            sans_dns=[host_mapping[1], self.charm.unit.name],
-        )
-
-        if not self.configure_certificates(sans):
+        if not self.setup_internal_certificates(
+            self.tls_manager.build_sans(
+                sans_ip=[self.state.unit.ip],
+                sans_dns=[self.state.unit.hostname, self.charm.unit.name],
+            )
```
marcoppenheimer, line 84: **suggestion:** This could be an `@property` of `tls_manager`, then we would call:

```python
if not self.setup_internal_certificates(self.tls_manager.sans)
```

Feels neater

src/events/cassandra.py
```diff
@@ -96,41 +94,147 @@ def _on_start(self, event: StartEvent) -> None:
...
             enable_peer_tls=self.state.unit.peer_tls.ready,
             enable_client_tls=self.state.unit.client_tls.ready,
             keystore_password=self.state.unit.keystore_password,
             truststore_password=self.state.unit.truststore_password,
         )
+        self.cluster_manager.prepare_shutdown()
+        self.charm.on[str(self.bootstrap_manager.name)].acquire_lock.emit()
 
+    def _start_subordinate(self, event: StartEvent) -> None:
+        if not self.state.cluster.is_active:
```
marcoppenheimer, line 144: **question:** Does `cluster.is_active` evaluate to `True` only after `_start_leader()` has ran once?

src/events/cassandra.py
```diff
@@ -96,41 +94,147 @@ def _on_start(self, event: StartEvent) -> None:
...
+                    raise BadSecretError()
+            except SecretNotFoundError:
+                logger.error("Cannot find user-defined system users secret")
+                raise BadSecretError()
+            except ModelError as e:
+                logger.error(f"Error accessing user-defined system users secret: {e}")
+                raise BadSecretError()
+        if self.state.cluster.cassandra_password_secret:
+            return self.state.cluster.cassandra_password_secret
+        return self.workload.generate_password()
```
marcoppenheimer: **todo:** This is a lot of nesting, I think it can be compartmentalized also.

First, let's move the `try: except` block somewhere else. Maybe some `events/secrets.py`, where we can handle secrets independently from the core bootstrap logic of this file.

In that file, we can also handle the boolean check for `system-users` by early-returning an empty string.

Lastly, we should put the `cassandra-password` as part of the cluster state.

So ultimately, this function would be removed entirely, and if we want to check|generate the password somewhere, we simply do:

```python
cassandra_password = self.state.cluster.cassandra_password or self.workload.generate_password()
```

src/events/cassandra.py
```diff
@@ -96,41 +94,147 @@ def _on_start(self, event: StartEvent) -> None:
...
+            cassandra_config_changed = self.config_manager.render_cassandra_config()
+            if env_changed or cassandra_config_changed:
+                self.cluster_manager.prepare_shutdown()
+                self.charm.on[str(self.bootstrap_manager.name)].acquire_lock.emit()
         except ValidationError as e:
             logger.debug(f"Config haven't passed validation: {e}")
             return
+        except BadSecretError:
+            event.defer()
+            return
```
marcoppenheimer, line 190: **todo:** I don't think we should ever defer for this, because it seems like it's only called when there's some Juju error (like can't reach the model or something). Let's just let it raise.

src/events/cassandra.py
```diff
@@ -96,41 +94,147 @@ def _on_start(self, event: StartEvent) -> None:
...
-            self.state.unit.peer_tls.rotation = False
-            self.state.unit.client_tls.rotation = False
+    def _on_secret_changed(self, event: SecretChangedEvent) -> None:
+        if not self.charm.unit.is_leader():
+            return
 
-        if self.state.unit.workload_state == UnitWorkloadState.ACTIVE:
-            self.charm.on[str(self.bootstrap_manager.name)].acquire_lock.emit()
+        if self.state.unit.workload_state == UnitWorkloadState.INSTALLING:
+            return
```
marcoppenheimer, line 197: **question:** What happens when we get `secret-changed`, but we're installing? Do we miss this event forever? Or after we set not `INSTALLING` does the password get read/created?

src/events/cassandra.py
```diff
@@ -144,17 +248,19 @@ def _on_update_status(self, _: UpdateStatusEvent) -> None:
...
             self.charm.on[str(self.bootstrap_manager.name)].acquire_lock.emit()
 
     def _on_collect_unit_status(self, event: CollectStatusEvent) -> None:
         try:
             self.charm.config
+            self._acquire_cassandra_password()
         except ValidationError:
             event.add_status(Status.INVALID_CONFIG.value)
+        except BadSecretError:
+            event.add_status(Status.INVALID_SYSTEM_USERS_SECRET.value)
```
marcoppenheimer, line 241: **note:** Same as earlier, I'd much rather we just `raise` than handle this everywhere. So here and below, we can just remove the check for the password and let `structured_config` handle it.

src/events/tls.py
```diff
@@ -47,20 +50,24 @@ def __init__(
...
         self.state = state
         self.workload = workload
+        self.config_manager = config_manager
+        self.bootstrap_manager = bootstrap_manager
         self.tls_manager = tls_manager
-        self.configure_certificates = configure_certificates
+        self.setup_internal_certificates = setup_internal_certificates
 
-        host_mapping = cluster_manager.network_address()
+        ip, hostname = cluster_manager.network_address()
```
marcoppenheimer, line 67: **praise:** This is tidy 👍🏾

src/managers/config.py
```diff
@@ -193,3 +169,133 @@ def _env_jmx_exporter_config(
...
+                    "store_type": "JKS",
+                    "protocol": "TLS",
+                }
+            }
+            if enabled
+            else {}
+        )
+
+    @staticmethod
+    def _cassandra_authentication_config(enabled: bool) -> dict[str, Any]:
```
marcoppenheimer, line 214: **nitpick:** It feels odd to have a private `@staticmethod`. Not important, just odd.

src/managers/config.py
```diff
@@ -167,11 +139,15 @@ def _map_env(env: Iterable[str]) -> dict[str, str]:
...
+
+    @staticmethod
+    def _merge_dicts(values: Iterable[dict[Any, Any]]) -> dict[Any, Any]:
         res = {}
-        for env in envs:
-            res.update(env)
-        return "\n".join([f"{key}={value}" for key, value in res.items()])
+        for value in values:
+            res.update(value)
+        return res
```
marcoppenheimer: **question:** Is this needed?

From an example that I know of which also uses YAML config, we set the dicts https://github.com/canonical/kafka-ui-operator/blob/3d71ad067cb493053187175dcc4482b4c5c74e66/src/managers/config.py#L132-L165.

Then we can just use default python dict merge syntax before formatting as a yaml string https://github.com/canonical/kafka-ui-operator/blob/3d71ad067cb493053187175dcc4482b4c5c74e66/src/managers/config.py#L190-L204

That feels much cleaner

marcoppenheimer (changes requested):

src/charm.py
marcoppenheimer: I now remembered that the `cassandra` user is the default.
After we've finished with cluster init, are we able to change that default user to `operator` in Cassandra? Or is it hard-coded, 'MUST be there' sort of thing?

src/core/statuses.py
welpaolo: @marcoppenheimer This is the expected behavior in spec DA146

welpaolo (approved):
Looks good on my side. After the resolution of  Marc's comments we should be good to go!

src/core/statuses.py
marcoppenheimer: Really? hmm, I guess I might need to change Kafka then.

src/core/config.py
Zvirovyi: > \- Mom, can we have **validator**?
> \- No, we have **validator** at home.
> \- Validator at home:
> image
> just a joke :)


Seriously speaking, just setting `type: secret` in charmcraft.yaml already enables the same validator, but provided by the Juju itself (Kafka has `type: string`):

image

src/charm.py
Zvirovyi: Okay, I'm done with changing `cassandra` user to `operator`.
But there already was a discussion on `src/literals.py`: https://github.com/canonical/cassandra-operator/pull/1#discussion_r2157670489. So, if we really want to move it to the constant, where we should put it (core / common / database manager)?

src/charm.py
Zvirovyi: This is an incorrect suggestion, as current code always puts `rotation` to `False`, while your code actually inverts it.
Originally, these `if` checks were added to mitigate warnings about missing TLS relation, because `peer_tls` or `client_tls` relations can be established after the peer relation, and writing `rotation` can trigger them. So, let's maybe leave them as-is?

src/core/state.py
Zvirovyi: Cassandra internal replication doesn't utilize any users, so `replication` cannot be implemented.
But with `operator` - done.

src/core/statuses.py
Zvirovyi: Continuing [our previous discussion on this](https://github.com/canonical/cassandra-operator/pull/8#discussion_r2285100829), from [DA146](https://docs.google.com/document/d/1uguJxAeNfUVl0gKhUzr-t-E_gMEfrbUELMjHY4vsBYw/edit?usp=sharing):

> If the secret does not contain a correct password or user name, the charm will be blocked and display a message asking the user for a valid secret content.

src/events/cassandra.py
Zvirovyi: We should return to [our previous huge thread on this](https://github.com/canonical/cassandra-operator/pull/3#discussion_r2178479053), and specifically to @deusebio clarification:

> The pattern that suggests @marcoppenheimer above does this, meaning that if cluster_name is something that has just been defined in the hook, the manager call's in the hook should provide that when calling render_cassandra_config.

Therefore we cannot do this currently, as `self.state.unit.ip` and `self.state.hostname` are changed in this exact hook.

src/events/cassandra.py
Zvirovyi: `cluster.is_active` relies on `cluster.state == ClusterState.ACTIVE`, that is set in `charm.py::bootstrap` function after the leader node is actually started.

src/events/cassandra.py
Zvirovyi: While we can move secret changed event handler into `events/secrets.py`, this code stays integrated in start and config changed events, and I don't see a way we can separate this logic from the secret changed event. And even if we want to share a part of the implementation, it becomes continuation of [our previous **unresolved** discussion on the similar case](https://github.com/canonical/cassandra-operator/pull/9#discussion_r2220009461).

Also, I'm not sure that this is what you've meant, but:
1. syncing 2 `event/` handlers through the state will be painful;
2. I'm against putting any logic into the state.

Also, from [your message below](https://github.com/canonical/cassandra-operator/pull/8#discussion_r2291812979), are you suggesting putting secret accessing / validation logic into `structured_config`?

src/events/cassandra.py
Zvirovyi: `INSTALLING` workload state means that this specific unit hasn't reached / completed start event yet. And start event reconciles the whole Cassandra configuration, including password from `system-users` secret, so it's safe to return early here.

src/events/cassandra.py
Zvirovyi: Do you mean putting secret accessing / validation logic into `structured_config`?

src/managers/config.py
Zvirovyi: If we don't need `self` here, why wouldn't make it `@staticmethod`?

src/managers/config.py
Zvirovyi: Nice catch!

src/charm.py
marcoppenheimer: Yeah I think we should move it somewhere common, because over the last 3 years the 'admin' user has had a different name at least 4 times, it's definitely going to change again.

If we aren't using `literals/constants`, maybe `core.models.ADMIN_USER=operator`? What do you think? I don't mind where exactly.

src/charm.py
marcoppenheimer: Oh I'm a fool, you're right, bad suggestion. Yeah let's leave it.

src/core/config.py
marcoppenheimer: OH REALLY? TIL, thank you that's much better 0 I'd much rather use the default Juju validator, OK, nevermind.

src/events/cassandra.py
marcoppenheimer: Good points, I'll try to be a bit more succinct.

This block does 3 different things:
1. Load the secret while handling Juju model errors
2. Checks if a password already exists in `state`
3. Generates a password

I think 1. should not be in `events/cassandra`. It's almost 'Juju-y business logic', and should be abstracted away somewhere else.

2. and 3. being here are 'event logic' for Cassandra start-up, so it's good to keep here.

src/events/cassandra.py
marcoppenheimer: Not quite. I think this comment is about `raise` vs `BlockedStatus`, which you rightly said we should `BlockedStatus` for this:

> If the secret does not contain a correct password or user name, the charm will be blocked and display a message asking the user for a valid secret content.

marcoppenheimer: As for where we do the secret accessing + validating, in https://github.com/canonical/cassandra-operator/pull/8#discussion_r2291802475 I suggested `events/secrets`. Maybe it would be better in `managers/secrets`? I don't really mind where it goes, but it's a lot of logic that isn't telling a charm dev 'what steps does cassandra need to start-up'.

src/managers/config.py
marcoppenheimer: I guess! Just seems odd but no harm.

src/events/cassandra.py
Zvirovyi: This error also indicates a secret validation error, so it's not only about ModelError during accessing the secret. But, let's drift from this case:
- [Kafka doesn't re-raise model errors](https://github.com/canonical/kafka-k8s-operator/blob/74530e20a0b61905f77e7388929f96d449de5931/src/events/user_secrets.py#L104C1-L106C22);
- Charmed etcd [sometimes does](https://github.com/canonical/charmed-etcd-operator/blob/57ec4100a885dcadf99e5425222789888188268b/src/events/etcd.py#L578-L586) and [sometimes doesn't](https://github.com/canonical/charmed-etcd-operator/blob/57ec4100a885dcadf99e5425222789888188268b/src/events/etcd.py#L578-L586);
- [Kyubi does](https://github.com/canonical/kyuubi-k8s-operator/blob/1385da1e8171cf0ea406374f5befa99e5b2acb0e/src/core/domain.py#L466).

My point is: let's just do this one way or another, and not mix solutions. So, do I understand you correctly, that we shouldn't suppress ModelError?

Zvirovyi: Also, this defer allows us handle the next scenario:
1. user sets `system-users` config to secret, that wasn't granted for application;
2. charm observes `ModelError` (ERROR permission denied), and instead of raising an error, it puts itself into blocked status;
3. user grants this secret to the application;
4. deferred config changed event now handles this secret properly.

src/charm.py
Zvirovyi: Done. I've put it into `core.literals.CASSANDRA_ADMIN_USERNAME = "operator"`. Check if it fits.

src/events/cassandra.py
Zvirovyi: Re 1:

> Not quite. I think this comment is about `raise` vs `BlockedStatus`, which you rightly said we should `BlockedStatus` for this:
> 
> > If the secret does not contain a correct password or user name, the charm will be blocked and display a message asking the user for a valid secret content.

Re 2:

> As for where we do the secret accessing + validating, in [#8 (comment)](https://github.com/canonical/cassandra-operator/pull/8#discussion_r2291802475) I suggested `events/secrets`. Maybe it would be better in `managers/secrets`? I don't really mind where it goes, but it's a lot of logic that isn't telling a charm dev 'what steps does cassandra need to start-up'.

So, do I understand correctly, that thread is settled on **Re 1** and **Re 2** isn't related here and rather to [#8 (comment)](https://github.com/canonical/cassandra-operator/pull/8#discussion_r2291802475)?

Zvirovyi: Resolved by putting `read_auth_secret` func in charm.py.

src/events/cassandra.py
Zvirovyi: Resolved by putting `read_auth_secret` func in charm.py.

src/events/cassandra.py
Zvirovyi: Resolved by putting `read_auth_secret` func in charm.py.

marcoppenheimer (approved):

PR #9, opened 2025-07-14
[DPE-7602] TLS implementation

BON4:
# Pull Request: Add Internal TLS Support and Refactor Certificate Management

## Overview

This PR introduces comprehensive internal TLS (Transport Layer Security) support for the Cassandra Operator. It implements certificate management, keystore/truststore handling, and event-driven TLS rotation, along with extensive unit tests.

### 1. Internal TLS Support
- Implements full internal TLS for Cassandra clusters, including:
  - Automatic generation of CA, server, and client certificates.
  - Support for certificate rotation.
  - Management of keystore and truststore files for Java-based workloads.
  - Integration with Juju secrets for secure storage of sensitive material.

### 2. Certificate Management
- Adds a `TLSManager` class responsible for:
  - Generating and storing CA, server, and client certificates.
  - Creating and updating keystore/truststore files.
  - Importing, updating, and removing certificates in the truststore.

### 3. Event-Driven TLS
- Introduces new events and event handlers for:
  - Certificate availability and rotation.
  - Relation changes that require TLS reconfiguration.
  - Automatic reconfiguration and restart of workloads when certificates change.

## Motivation

- Secure Cassandra clusters with automated, robust TLS.
- Simplify certificate management for operators and users.
- Improve maintainability and test coverage of the codebase.

## Testing

- All new features are covered by unit tests.
- Manual integration testing performed for multi-node and rolling upgrade scenarios.


BON4:
Cassandra snap modification needed. So waiting for it.
https://github.com/canonical/charmed-cassandra-snap/pull/14#issue-3231961408

src/managers/config.py
```diff
@@ -61,6 +65,7 @@ def render_cassandra_config(
             "inter_dc_tcp_nodelay": False,
             "internode_compression": "dc",
             "listen_address": listen_address or self.listen_address,
+            "broadcast_address": listen_address or self.listen_address,
```
Zvirovyi: Why should we setup `broadcast_address`?

src/managers/config.py
```diff
@@ -34,6 +34,10 @@ def render_cassandra_config(
         cluster_name: str | None = None,
         listen_address: str | None = None,
         seeds: list[str] | None = None,
+        enable_peer_tls: bool = False,
+        enable_client_tls: bool = False,
+        keystore_password: str | None = None,
+        truststore_password: str | None = None,
```
Zvirovyi, line 49: Those values should be included in initializer to provide a default value, that can be specifically overridden in this method. We already passthrough "default" values [here](https://github.com/canonical/cassandra-operator/blob/f73fe82874205dc75913b06153a6b96a1956291b/src/charm.py#L39). I recommend you to check our [previous discussion on this decision](https://github.com/canonical/cassandra-operator/pull/3#discussion_r2178479053).

src/managers/cluster.py
```diff
@@ -31,6 +30,15 @@ def is_healthy(self) -> bool:
         except ExecError:
             return False
 
+    @property
+    def is_node_in_ring(self) -> bool:
```
Zvirovyi: Is this method used anywhere? I prefer to not drag unused code to project even though it seems helpful. Can this error be occurred with our new rolling ops bootstrap mechanism?

src/core/workload.py
```diff
@@ -92,6 +117,35 @@ def path_exists(self, path: str) -> bool:
...
+
+    @abstractmethod
+    def client_tls_ready(self) -> bool:
+        """Whether Cassandra Client TLS files are ready."""
         pass
+
+    @property
+    @abstractmethod
+    def installed(self) -> bool:
+        """Checks whether the workload service is installed."""
```
Zvirovyi: I don't think we need this method and any further checks based on it as `install` event will be called before `start` or any `relation-joined` events, so we can assume workload is installed all the times. And even if installation wasn't successful, an error will be raised and charm will emerge corresponding status. Correct me if I'm wrong.

src/workload.py
```diff
@@ -130,3 +138,28 @@ def exec(self, command: list[str], suppress_error_log: bool = False) -> tuple[st
...
+
+    @override
+    def client_tls_ready(self) -> bool:
+        return all(
+            self.path_exists(f.as_posix())
+            for f in [
+                self.cassandra_paths.client_truststore,
+                self.cassandra_paths.client_keystore,
+            ]
+        )
```
Zvirovyi: Let's not bring any logics into workload. Can this be moved to the manager, especially when we can use `self.cassandra_paths.client_truststore.exists()`?

src/events/tls.py
```diff
@@ -0,0 +1,270 @@
+"""TLS event handling and related logic for the charm."""
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
```
Zvirovyi:
```suggestion
#!/usr/bin/env python3
# Copyright 2025 Canonical Ltd.
# See LICENSE file for licensing details.

"""TLS event handling and related logic for the charm."""
```

src/managers/tls.py
```diff
@@ -0,0 +1,521 @@
+# Copyright 2025 Canonical Ltd.
```
Zvirovyi, line 2:
```suggestion
#!/usr/bin/env python3
# Copyright 2025 Canonical Ltd.
```

src/core/state.py
```diff
@@ -74,6 +118,171 @@ def _field_setter_wrapper(self, field: str, value: str) -> None:
...
+
+    def setup_provider_certificates(self, crts: ProviderCertificate) -> None:
+        """Set up certificate, CSR, and chain from a ProviderCertificate object.
+
+        Args:
+            crts: ProviderCertificate object containing certificate, CSR, and chain.
+        """
+        self.certificate = crts.certificate
+        self.csr = crts.certificate_signing_request
+        self.chain = crts.chain
```
Zvirovyi: Considering that this is just a proxy helper for 3 setters, and the only 2 usages I can find is literally alongside other `peer_tls` setters, it will be just better to not use it in the first place and keep things simple.

### Example on what do I mean

Instead of:

```python
self.state.unit.peer_tls.setup_provider_certificates(provider_crt)
self.state.unit.peer_tls.private_key = pk
self.state.unit.peer_tls.ca = self.state.cluster.internal_ca
```

We can just do:

```python
self.state.unit.peer_tls.certificate = provider_crt.certificate
self.state.unit.peer_tls.csr = provider_crt.certificate_signing_reques
self.state.unit.peer_tls.chain = provider_crt.chain
self.state.unit.peer_tls.private_key = pk
self.state.unit.peer_tls.ca = self.state.cluster.internal_ca
```

Anyways it will be a place for the several setters, so it doesn't make sense to passthrough a part of it (especially when it's literally 3 of 5 rows with only 2 occurrences) through this helper.

src/core/state.py
```diff
@@ -74,6 +118,171 @@ def _field_setter_wrapper(self, field: str, value: str) -> None:
...
+
+    @property
+    def private_key(self) -> Optional[PrivateKey]:
```
Zvirovyi: Let's not introduce `Optional[...]` when we already using `... | None` type hint everywhere.

src/core/state.py
```diff
@@ -136,6 +345,57 @@ def workload_state(self) -> UnitWorkloadState:
...
+        """TLS state for internal (peer) communications."""
+        return TLSContext(self.relation, self.data_interface, self.unit, TLSScope.PEER)
+
+    @property
+    def client_tls(self) -> TLSContext:
+        """TLS state for external (client) communications."""
+        return TLSContext(self.relation, self.data_interface, self.unit, TLSScope.CLIENT)
+
+    @property
+    def keystore_password(self) -> str:
```
Zvirovyi, line 349: @deusebio @marcoppenheimer @welpaolo 
I've already heard several times about inconvenience about "hidden secrets logic", and even got personally confused multiple times when I worked on the PostgreSQL & Opensearch operators.
I have an idea of naming all of the state fields with `_secret` suffix if it really goes into secret. It will not hurt productivity in any means, but will significantly simplify understanding of Cassandra operator state, especially for newbie users. Examples: `cassandra_password_secret`, `keystore_password_secret`. What do you think?

src/core/workload.py
```diff
@@ -18,12 +22,33 @@ class CassandraPaths:
...
+
+    @property
+    def client_truststore(self) -> pathops.PathProtocol:
+        """TODO."""
+        return self.tls_dir / f"{TLSScope.CLIENT.value}-truststore.jks"
+
+    @property
+    def client_keystore(self) -> pathops.PathProtocol:
+        """TODO."""
+        return self.tls_dir / f"{TLSScope.CLIENT.value}-keystore.p12"
```
Zvirovyi: I think if we unmark this as property and provide an TLSScope argument like `def get_client_keystore(self, scope: TLSScope) -> ...`, we can then reuse these paths in the TLS manager so we wouldn't need to manually dereferencing them like `(self.workload.cassandra_paths.tls_dir / f"{scope.value}-truststore.jks").as_posix()` and instead do `self.workload.cassandra_paths.get_client_keystore(scope).posix()`.

src/managers/tls.py
```diff
@@ -0,0 +1,521 @@
...
+            tls_state: The resolved TLS context containing all required artifacts.
+            keystore_password: Password for the keystore.
+            trust_password: Password for the truststore.
+        """
+        scope = tls_state.scope
+        self.set_private_key(tls_state.private_key, scope)
+        self.set_ca(tls_state.ca, scope)
+        self.set_bundle(tls_state.bundle, scope)
+        self.set_chain(tls_state.chain, scope)
+        self.set_certificate(tls_state.certificate, scope)
```
Zvirovyi, line 377: These functions actually don't use `scope` parameter. It should either be fixed or argument should be removed.

src/core/state.py
```diff
@@ -74,6 +118,171 @@ def _field_setter_wrapper(self, field: str, value: str) -> None:
...
+        if not self.certificate or not self.private_key or not self.ca:
+            raise RuntimeError("TLS state is incomplete")
+        return ResolvedTLSContext(
+            private_key=self.private_key,
+            ca=self.ca,
+            certificate=self.certificate,
+            chain=self.chain,
+            bundle=self.bundle,
+            scope=self.scope,
+        )
```
Zvirovyi, line 272: Maybe we can get rid of `ResolvedTLSContext` and just pass `TLSContext` to TLS manager functions?

src/events/cassandra.py
```diff
@@ -80,19 +92,30 @@ def _on_start(self, event: StartEvent) -> None:
             return
 
         self.config_manager.render_cassandra_config(
-            cluster_name=self.state.cluster.cluster_name,
+            cluster_name=self.charm.config.cluster_name,
```
Zvirovyi: No, leave it like it was.

src/managers/tls.py
```diff
@@ -0,0 +1,521 @@
...
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes
+
+from common.exceptions import ExecError
+from core.state import ResolvedTLSContext, TLSScope
+from core.workload import WorkloadBase
+
+logger = logging.getLogger(__name__)
+
+Sans = TypedDict("Sans", {"sans_ip": list[str], "sans_dns": list[str]})
```
Zvirovyi: Is there any reason behind `TypedDict`? I don't have a strong opinion on that, but can we use `dataclass` for this? Just to not create any redundant dictionaries.
@deusebio @marcoppenheimer @welpaolo, what do you think?

Zvirovyi (changes requested):
Can you provide more information - is TLS becomes mandatory for Cassandra? If so, when there's no `certificate` relation established, does charm generate temporary internal-only credentials?
Also, please, fix docstrings through your PR. I found it missing / wrong in multiple places here.
Will review tests after most of the raised questions are resolved.
Thanks for the work!

src/core/state.py
deusebio: Yes, sure. I don't think I have concern with this.

src/core/state.py
deusebio: I personally like the attempt to use a more pure dependency-free dataclass to pass things around. `ResolvedTLSContext` does not depend on `ops` and it can easily be instantiated outside of charms to run unit and integration tests. I've always seen the `ops` second order dependency into managers (since managers depends on the context that depends on ops) a bit dirty

Although I undestand it is not exactly how we do things on other charms, I would personally give it a go to experiment the pattern. It should not introduce very large deviation anyway

deusebio: **niptick** the `resolved` method could probably be a `property`

src/core/workload.py
deusebio: I would be happy either way. I don't have any strong option for or against ether option

src/managers/tls.py
deusebio: Yes, I would prefer a dataclass too honestly

src/managers/tls.py
```diff
@@ -0,0 +1,521 @@
...
+        )
+
+        logger.debug(
+            f"Certificate sans ip: {certificate.sans_ip}, \
+                       sans dns: {certificate.sans_dns}"
+        )
+
+        return provider_cert, unit_key
+
+    def set_ca(self, ca: Certificate, scope: TLSScope) -> None:
```
deusebio, line 169: **praise** I personally like that you are using `set_something(value: type)` api. I believe in other charms we just use the `set_something()` with no arguments, with the information figured out from the context implicitly, which however makes the code a bit more obscure. This is a lot clearer.

src/workload.py
deusebio: Yes, I believe @Zvirovyi has a point. This start to seem a bit outside of the scope for the workload, and more relevant to the manager.

src/events/cassandra.py
```diff
@@ -160,3 +199,55 @@ def _update_network_address(self) -> bool:
             and old_hostname is not None
             and (old_ip != self.state.unit.ip or old_hostname != self.state.unit.hostname)
         )
+
+    def _check_and_set_certificates(self) -> bool:
```
deusebio: **suggestion** the part here seems rather similar to the login on `tls_broken_events`. I'm thinking if there are ways to have a central method that can be referenced in both places, to avoid copy and paste. 

if we factor out this method, some properties may become arguments (e.g. sans). You could also use a bit of functional programming paradigms, like:

```
class CassandraEvents(Object):
    
    def configure_certificates(self, sans: Sans) -> bool:
        ... 

class TLSEvents(Object):
    """Manage TLS-related events and event sources for the charm."""

    refresh_tls_certificates = EventSource(RefreshTLSCertificatesEvent)

    def __init__(
        self,
        ...,
        initialize_certificates: Callable[[Sans], bool]
    ):
        ...
   
# Then on charm.py

self.cassandra_events = CassandraEvents(
       self,
       state=self.state,
       workload=self.workload,
       cluster_manager=self.cluster_manager,
       config_manager=config_manager,
       bootstrap_manager=bootstrap_manager,
       tls_manager=self.tls_manager,
)

self.tls_events = TLSEvents(
    self,
    state=self.state,
    workload=self.workload,
    cluster_manager=self.cluster_manager,
    tls_manager=self.tls_manager,
    initialize_certificates = self.cassandra_events.configure_certificates
)
```

tests/integration/test_tls.py
```diff
@@ -0,0 +1,270 @@
...
+KEYSTORE_PASSWORD = "keystore-password"
+CLIENT_CA_CERT = "client-ca-cert"
+PEER_CA_CERT = "peer-ca-cert"
+
+TLS_NAME = "self-signed-certificates"
+
+PEER_PORT = 7000
+CLIENT_PORT = 9042
+
+
```
deusebio, line 25: **praise** well written tests with the checks of the various ux and combinations

deusebio (commented):
The work looks already quite good, and I don't have many concerns with the code here. I'd just like someone that has just worked with the `TLSv4` library to review the logic. I believe @theoctober19th can really help here. Please Bikalpa have a closer look to the `events/tls.py` where v4 is used. Thanks!

src/core/state.py
theoctober19th: +1

src/core/state.py
```diff
@@ -74,6 +118,171 @@ def _field_setter_wrapper(self, field: str, value: str) -> None:
...
+        return PrivateKey.from_string(raw)
+
+    @private_key.setter
+    def private_key(self, value: PrivateKey | None) -> None:
+        if not value:
+            return self._field_setter_wrapper(f"{self.scope.value}-private-key", "")
+        self._field_setter_wrapper(f"{self.scope.value}-private-key", value.raw)
+
+    @property
+    def csr(self) -> Optional[CertificateSigningRequest]:
```
theoctober19th: Similar to the comment above, let's use `CertificateSigningRequest | None` in place of `Optional`.

src/core/state.py
theoctober19th: +1, let's do that.

src/events/cassandra.py
Zvirovyi: @deusebio
I don't have a strong opinion on the Callable passthrough, but if we want to have a shared code, let's then just put it inside `charm.py` or some _common place_ and not a particular `...Events` object, just like we already do with the boostrap. Because otherwise it raises question on "what `...Events` object should be an origin of the function?" (now it doesn't make sense to put TLS-related function in **general** `...Events` object) and really breaks separation of the responsibilities.

Also, just as an idea, can we have multiple listeners `on_start` and just sync them?

theoctober19th (commented):
I had initial look at the `tls/events` file as requested by @deusebio and it looks good to me. One thing I noticed Cassandra operator does differently from other charms we have (like Kyuubi) is that when TLS relation is dropped, the charm defaults back to internal TLS and thus will always have TLS enabled someway or the other.  I'll come back to the PR later for a more detailed review.

src/core/workload.py
BON4:
```python
    def _on_peer_certificate_available(self, event: CertificateAvailableEvent) -> None:
        """Handle `certificate_available` event.

        After provider updates signed certs for peer TLS relation.
        """
        if not self.state.peer_relation:
            logger.warning("No peer relation on certificate available")
            event.defer()
            return

        if not self.workload.installed:
            logger.warning("Workload is not yet installed")
            event.defer()
            return
```

I thought `_on_peer_certificate_available` could be called before workload is installed, in this case there would be an error, and call wont be deferred?

src/workload.py
BON4: Moved to tls_manager. Remained as property just becouse it would be inconvenient to use it like: `self.cassandra_paths.client_truststore.exists() and self.cassandra_paths.client_keystore.exists()`

src/events/cassandra.py
BON4: I moved `configure_internal_certificates` to `charm.py`, as you suggested, and now pass it explicitly through constructors instead of keeping it inside any `...Events` object. This keeps the shared logic in a common place and avoids mixing responsibilities between event handlers.

BON4: @marcoppenheimer

deusebio (approved):
The code looks good and the tests shows that it does things as expected. I don't have any concern with this. It is fine to be merged. 

To the point from @theoctober19th , yes the current ux here is the one expected for the Cassandra charm, which align with the implementations on etcd and that also Kafka has recently been updating/implementing. So all good!

welpaolo (approved):
Great work! It is really solid based on the state of the charm!
As future work, when scaling feature is available I will also suggest to validate the tls feature on scaling up and down and also when the a unit is killed.

theoctober19th (approved):

PR #13, opened 2025-08-29
Seed node management & Full cluster restart handling

Zvirovyi:
# Overview

This PR is based off [DA213 - Cassandra single seed node management specification](https://docs.google.com/document/d/1saFsnZFViogeyFGG8JFQJAfooxs-tceZo8Estz_WBQQ/edit?usp=sharing), that is currently on-review. It adds new peer changed & leader event handlers and more robust checks in start event handler to ensure cluster successfully recovers itself after full cluster restart or leader change.

# Cassandra seed node <> Juju leader desync

While currently Cassandra operator always configures Juju leader as Cassandra seed node, this PR introduces possibility of seed node setting being assigned to the another unit. It allows cluster to reliably recover from the full cluster restart as seed node will be pointing to knowingly healthy unit. From the future-proof perspective, with the multiple seed node management (that is planned) it will be impossible to hold a sync of seed node with Juju leader, therefore this work is done in the right direction. It also worth mentioning, that the next full cluster config change (TLS change, for example) will again assign seed node to the leader unit to ensure right restart order.

### DIFF
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -299,7 +299,6 @@ class CassandraEvents(Object):
         self._recover_seeds(event)
 
     def _on_update_status(self, _: UpdateStatusEvent) -> None:
-        # TODO: add peer relation change hook for subordinates to update leader address too
         if (
             self._update_network_address()
             and self.state.unit.workload_state == UnitWorkloadState.ACTIVE
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -122,7 +122,7 @@ class CassandraEvents(Object):
         ):
             self.state.cluster.seeds = [self.state.unit.peer_url]
 
-        if not self._check_seeds():
+        if not self._are_seeds_reachable:
             logger.debug("Deferring leader on_start due to seeds not being ready")
             event.defer()
             return
@@ -184,7 +184,7 @@ class CassandraEvents(Object):
             event.defer()
             return
 
-        if not self._check_seeds():
+        if not self._are_seeds_reachable:
             logger.debug("Deferring subordinate on_start due to seeds not being ready")
             event.defer()
             return
@@ -368,7 +368,8 @@ class CassandraEvents(Object):
 
         event.add_status(Status.ACTIVE.value)
 
-    def _check_seeds(self) -> bool:
+    @property
+    def _are_seeds_reachable(self) -> bool:
         return self.state.unit.peer_url in self.state.cluster.seeds or any(
             unit.workload_state == UnitWorkloadState.ACTIVE
             and DatabaseManager(
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -28,7 +28,7 @@ from tenacity import Retrying, stop_after_delay, wait_exponential, wait_fixed
 from common.exceptions import BadSecretError
 from core.config import CharmConfig
 from core.literals import CASSANDRA_ADMIN_USERNAME
-from core.state import PEER_RELATION, ApplicationState, ClusterState, UnitWorkloadState
+from core.state import PEER_RELATION, ApplicationState, UnitWorkloadState
 from core.statuses import Status
 from core.workload import WorkloadBase
 from managers.cluster import ClusterManager
@@ -88,6 +88,9 @@ class CassandraEvents(Object):
     def _on_start(self, event: StartEvent) -> None:
         self._update_network_address()
 
+        if self.state.unit.workload_state == UnitWorkloadState.ACTIVE:
+            self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
+
         if not self.setup_internal_certificates(
             self.tls_manager.build_sans(
                 sans_ip=[self.state.unit.ip],
@@ -112,32 +115,12 @@ class CassandraEvents(Object):
             self._start_subordinate(event)
 
     def _start_leader(self, event: StartEvent) -> None:
-        if self.state.cluster.is_active:
-            self.state.cluster.state = ClusterState.RECOVERING
-            self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
-            event.defer()
-            return
-
-        if self.state.cluster.state == ClusterState.RECOVERING:
-            if any(
-                unit.workload_state != UnitWorkloadState.WAITING_FOR_START
-                for unit in self.state.other_units
-            ):
-                event.defer()
-                return
-            else:
-                self.state.cluster.state = ClusterState.UNKNOWN
-
         if not self.state.cluster.seeds or all(
             self.state.cluster.seeds[0] != unit.peer_url for unit in self.state.units
         ):
             self.state.cluster.seeds = [self.state.unit.peer_url]
 
-        if self.state.cluster.seeds[0] != self.state.unit.peer_url and any(
-            unit.workload_state != UnitWorkloadState.ACTIVE
-            for unit in self.state.other_units
-            if unit.peer_url in self.state.cluster.seeds
-        ):
+        if not self._check_seeds():
             event.defer()
             return
 
@@ -185,25 +168,13 @@ class CassandraEvents(Object):
         self.cluster_manager.prepare_shutdown()
 
     def _start_subordinate(self, event: StartEvent) -> None:
-        if (
-            self.state.cluster.is_active
-            and self.state.unit.workload_state == UnitWorkloadState.ACTIVE
-        ):
-            event.defer()
-            return
-
-        if self.state.cluster.state == ClusterState.RECOVERING:
-            if self.state.unit.workload_state != UnitWorkloadState.WAITING_FOR_START:
-                self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
-            event.defer()
-            return
-
-        if (
-            not self.state.cluster.is_active
-            and self.state.unit.peer_url not in self.state.cluster.seeds
-        ):
+        if not self.state.cluster.is_active:
             self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
             logger.debug("Deferring subordinate on_start due to cluster isn't initialized yet")
+            event.defer()
+            return
+
+        if not self._check_seeds():
             event.defer()
             return
 
@@ -370,6 +341,25 @@ class CassandraEvents(Object):
             event.add_status(Status.INVALID_SYSTEM_USERS_SECRET.value)
 
         event.add_status(Status.ACTIVE.value)
+
+    def _is_seed_active(self, host: str) -> bool:
+        return DatabaseManager(
+            hosts=[host],
+            user=CASSANDRA_ADMIN_USERNAME,
+            password=self.state.cluster.operator_password_secret,
+        ).check()
+
+    def _check_seeds(self) -> bool:
+        return self.state.unit.peer_url in self.state.cluster.seeds or any(
+            unit.workload_state == UnitWorkloadState.ACTIVE
+            and DatabaseManager(
+                hosts=[unit.ip],
+                user=CASSANDRA_ADMIN_USERNAME,
+                password=self.state.cluster.operator_password_secret,
+            ).check()
+            for unit in self.state.other_units
+            if unit.peer_url in self.state.cluster.seeds
+        )
 
     def _recover_seeds(self, event: EventBase) -> None:
         if self.state.unit.workload_state == UnitWorkloadState.INSTALLING:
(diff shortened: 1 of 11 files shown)


tests/integration/test_full_cluster_restart.py
```diff
@@ -0,0 +1,47 @@
...
+        num_units=3,
+    )
+    juju.wait(
+        ready=lambda status: jubilant.all_agents_idle(status) and jubilant.all_active(status),
+        delay=20,
+        timeout=1000,
+    )
+
+
+def test_lxc_restart(juju: jubilant.Juju) -> None:
```
deusebio, line 29: This tests is good, but could we also have some tests that are aligned with other charms, e.g. Kafka, see [here](https://github.com/canonical/kafka-operator/blob/main/tests/integration/ha/test_ha.py#L334). Have we taken the `lxc restart --all` scenario from some other charms? If so, we could probably have both, to make sure we are consistent

src/events/cassandra.py
```diff
@@ -281,6 +368,34 @@ def _on_collect_app_status(self, event: CollectStatusEvent) -> None:
 
         event.add_status(Status.ACTIVE.value)
 
+    def _check_seeds(self) -> bool:
```
deusebio: **todo** could you add some comment on what is being checked? Or maybe rename the function into `is_seed_reachable` which is a bit more clear on what we are checking

src/events/cassandra.py
```diff
@@ -77,12 +88,17 @@ def _on_install(self, _: InstallEvent) -> None:
     def _on_start(self, event: StartEvent) -> None:
         self._update_network_address()
 
+        if self.state.unit.workload_state == UnitWorkloadState.ACTIVE:
```
deusebio, line 91: **question** When is this happening? I'm not sure I follow what's happening here.

src/events/cassandra.py
```diff
@@ -117,33 +159,33 @@ def _start_leader(self, event: StartEvent) -> None:
...
-            truststore_password=self.state.unit.truststore_password,
-        )
         self.cluster_manager.prepare_shutdown()
-        self.charm.on[str(self.bootstrap_manager.name)].acquire_lock.emit()
 
     def _start_subordinate(self, event: StartEvent) -> None:
-        if not self.state.cluster.is_active:
+        if (
+            not self.state.cluster.is_active
+            and self.state.unit.peer_url not in self.state.cluster.seeds
```
deusebio: **suggestion** I feel we could have a simple property/method `is_seed`, such that it becomes very readable 

`if not self.state.cluster.is_active and not self.state.unit.is_seed`

I feel this `peer_url is self.state.cluster.seeds` comes up rather frequency, for which a more readable expression could be beneficial

src/events/cassandra.py
```diff
@@ -101,13 +117,39 @@ def _on_start(self, event: StartEvent) -> None:
...
-        try:
-            self.state.cluster.operator_password_secret = self._acquire_operator_password()
-        except BadSecretError:
+        if not self.state.cluster.seeds or all(
+            self.state.cluster.seeds[0] != unit.peer_url for unit in self.state.units
+        ):
+            self.state.cluster.seeds = [self.state.unit.peer_url]
+
+        if not self._check_seeds():
+            logger.debug("Deferring leader on_start due to seeds not being ready")
```
deusebio, line 124: **question** For my understanding, could you clarify how this can be by-passed at the start? Since the leader should generally be the seed, how can check pass if the unit (that should be the seed) has not started yet. 

It obviously work, otherwise the tests would not pass, but I'm not sure I fully understand how this is possible

deusebio (commented):
The only major comment is that I'd like to make sure that the tests are consistent with the ones present in other charms. 

Apart from these I have few nipticks/suggestions, and some questions that will help me understand the workflow better. Thanks!

tests/integration/test_full_cluster_restart.py
marcoppenheimer: It's a new scenario, but it's (imo) better than the equivalent ones we have in other charms - it's a good test.

I'd like to see SIGTERM/KILL/STOP tests also though.

src/events/cassandra.py
marcoppenheimer: I think this is a `has_already_started` flag.

If we get `start` event again, it's probably something odd like a VM reboot

src/events/cassandra.py
```diff
@@ -101,13 +117,39 @@ def _on_start(self, event: StartEvent) -> None:
...
 
     def _start_leader(self, event: StartEvent) -> None:
-        self.state.cluster.seeds = [self.state.unit.peer_url]
-        try:
-            self.state.cluster.operator_password_secret = self._acquire_operator_password()
-        except BadSecretError:
+        if not self.state.cluster.seeds or all(
+            self.state.cluster.seeds[0] != unit.peer_url for unit in self.state.units
+        ):
+            self.state.cluster.seeds = [self.state.unit.peer_url]
```
marcoppenheimer: **todo:** Could this just be part of the `unit` state? `unit.is_seed: bool`?

src/events/cassandra.py
```diff
@@ -281,6 +368,34 @@ def _on_collect_app_status(self, event: CollectStatusEvent) -> None:
...
+        return self.state.unit.peer_url in self.state.cluster.seeds or any(
+            unit.workload_state == UnitWorkloadState.ACTIVE
+            and DatabaseManager(
+                hosts=[unit.ip],
+                user=CASSANDRA_ADMIN_USERNAME,
+                password=self.state.cluster.operator_password_secret,
+            ).check()
+            for unit in self.state.other_units
+            if unit.peer_url in self.state.cluster.seeds
+        )
```
marcoppenheimer: **todo:** This is really complicated, we should break it up in to sections.

First, as mentioned in the other comment, just have `unit.seed: bool` as part of the UnitContext.

Second, have `state.cluster.seeds: list[UnitContext]` as part of the cluster state.

Third, consider adding the `DatabaseManager.check()` as part of the `UnitContext`. It could look like this:
```python
class UnitContext:
    def __init__(self, ..., password: str | None = None) -> None:
        # password gets passed in ClusterContext when we initialise Units
        # as it'll always be a fixed `operator_password_secret`
        self._password = password

    @property
    def _manager(self) -> DatabaseManager:
        return DatabaseManager(
                hosts=[self.ip],
                user=CASSANDRA_ADMIN_USERNAME,
                password=self._password
        )
	
	@property
    def is_up(self) -> bool:
        return self._manager.check() 
```

Now, this code (and a lot of others in this file) looks like:

```python
def _check_seeds(self) -> bool:
    for unit in self.state.cluster.seeds:
        if unit == self.state.unit or unit.is_up:
            return True

    return False
```

Just reading that sentence completely explains what this code is doing.

marcoppenheimer: Ah, I see you're using `seeds` as a setter property.

We could be very clever here, and do something like:

```python
@seeds.setter
def seeds(self, value: list[UnitContext]):
    urls = [unit.peer_url for unit in value]
    self.relation_data(..., urls)  # I forgot the relation_data.set

@property
def seeds(self) -> list[UnitContext]:
    unit_map = {unit.peer_url: unit for unit in self.state.cluster.units}
    return [unit_map[seed_url] for seed_url in self.relation_data.get("seeds")]  # again, forgot the relation_data.get

Comment (marcoppenheimer) [reply src/events/cassandra.py]: tl;dr - put more things in Unit state. We should have a complex 'backend API' in state/managers, and very very simple 'frontend API' with events.

File: src/managers/database.py [hunk 8]
```diff
@@ -33,6 +40,27 @@ def __init__(self, hosts: list[str], user: str, password: str):
...
+                session.execute("SELECT release_version FROM system.local")
+                return True
+        except NoHostAvailable as e:
+            if e.errors:
+                for host_error in e.errors.values():
+                    if isinstance(host_error, AuthenticationFailed):
+                        return True
+            return False
+        except Exception:
+            return False
```
marcoppenheimer: **todo:** Small request to debug-log the exceptions here so it's not completely hidden

marcoppenheimer (commented):
Looks good! Left some early comments, mostly about compartmentalising the code elsewhere outside of `events`.

Agree with Enrico's request for a similar implementation with the `SIGxxxx` tests, but maybe that's coming later?

src/events/cassandra.py
Zvirovyi: Exactly! This will occur on VM reboot (e.g. `lxc restart --all`).

src/events/cassandra.py
Zvirovyi: There is check for it: `self.state.unit.peer_url in self.state.cluster.seeds` in `_are_seeds_reachable`. When unit is seed itself, it bypasses connectivity check.

src/events/cassandra.py
Zvirovyi: Also re to @marcoppenheimer:

> **todo:** Could this just be part of the `unit` state? `unit.is_seed: bool`?

`ApplicationState.unit` accepts `self.peer_unit_interface`, while `ApplicationState.cluster` - `self.peer_app_interface`. So, should we then break this separation and pass both interfaces to the `UnitState`, or should we place it somewhere else?

marcoppenheimer: In `ApplicationState`, when you initialise `unit` and `units`.

```python
    @property
    def unit(self) -> UnitContext:
        """This unit context."""
        return UnitContext(
            relation=self.peer_relation,
            data_interface=self.peer_unit_interface,
            component=self.model.unit,
            _seeds=self.cluster.seeds
        )
```

src/events/cassandra.py
Zvirovyi: @deusebio Should we start using managers in the application state? Doesn't this break separation of managers / state? From [our previous thread with you](https://github.com/canonical/cassandra-operator/pull/1#discussion_r2157673899):

> Most importantly, the current implementation breaks the dependency hierarchy
> 
> ```
> common < core < managers < handlers < root classes
> ```


Also, it makes me uncomfortable to have the general DatabaseManager, that is instantiated in charm.py (and reused in `...Events`), and also to have the same DatabaseManager, that is instantiated in the state for every unit. How then we should approach to functionality of the DatabaseManager? Should we somehow separate some of the methods to be used only through ApplicationState, and another to general instance? Or should we migrate to using only ApplicationState approach? Because I personally don't thing making multiple ways of accessing the same manager functionality will improve our codebase.

While `is_seed` and `def seeds -> set[UnitContext]` actually provide **different**, more user-friendly interface, `is_up` is just a proxy for DatabaseManager instantiation. Maybe then it will be better to have `check(host_override: str)` method directly in DatabaseManager? So we'll be able to check different host (with override argument) without manual instantiation of DatabaseManager inside `are_seeds_reachable` method.

@marcoppenheimer

tests/integration/test_full_cluster_restart.py
Zvirovyi: Let's do this in the separate ticket specifically for HA tests?

src/events/cassandra.py
deusebio: I believe Marc has a point on the readability, and I also found the codebase sometimes hard to read, where some higher-level functions or attributes of Unit can be rather beneficial to follow the logic. 

But I fully agree with @Zvirovyi that `is_up` should be a manager concern. Since it requires to connect to the platform, perform an API call, and parse its response, this is something I would expect from a manager (and that can also done outside of charms).

I'm not as uncomfortable to define the manager in `charm.py` or in the constructor for the cassandra Eventhandler though. That should be fine IMHO. There are also a couple of patterns one can use to handle the unit, like:
1. Adding `unit` argument to the check method to override the default one
2. Providing a function like:
```
def with_unit(self, unit_ip: str):
    return CassandraManger(hosts=[unit_ip], user=self.user, password=self.password)
```

Such that one can easily do `self.manager.with_unit(unit_ip).check()`. I believe that is more readable than the current 3 lines

deusebio (approved):
From a codebase perspective, it looks good to be merged to me (once the CI is happy). Thank you for replying to the various questions. 

Please have the another tickets to extend the testing. I'm fine to have it done in a separate PR

src/events/cassandra.py
marcoppenheimer: IMO, managers are completely independent of any 'charm' code. They should be safe to be instantiated anywhere, and are stateless, just initialising based on some input args.

marcoppenheimer: If the manager is for the 'cluster' rather than for a 'unit', that's fine to have it not be in `UnitContext`.

Zvirovyi: We agreed to leave database manager as-is.

tests/integration/test_full_cluster_restart.py
deusebio: https://warthogs.atlassian.net/browse/DPE-8322

PR #17, opened 2025-09-22
[DPE-7707] Improve bootstrapping process

Zvirovyi:
# Bootstrap changes

* Bootstrap process is now split into multiple event calls rather than being single long waiting operation.
* Charm now detects bootstrap failures such as service shutdown (due to exception) and unknown bootstrap state.
* When Cassandra node's bootstrap state is `PENDING`, charm tries to resume it with `nodetool bootstrap resume`.
* When charm detects bootstrap failure, it sets `CANT_START` workload state, during which the new configuration can be applied. Update status triggers retry of the bootstrapping.
* Shutdown preparement (`nodetool drain`) is moved from the event handlers to bootstrap process handler and now is based on whether workload is active rather than on the active workload state.

# Other changes

* `LockManager` is introduced as an replacement of RollingOpsManager, that provides simple `try_lock` & `release` API to manage unit-exclusive lock without ties to the specific event execution.
* `ClusterManager` renamed to `NodeManager`.
* `NodeManager` now relies on `UN` status of the node in `nodetool status` output.
* New `UnitContext` properties `is_ready` & `is_config_change_eligible` & `is_operational` were introduced to improve many events early exit & defer checks based on the state.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -112,6 +112,11 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
             event.defer()
             return
 
+        if not self.workload.is_alive():
+            self.state.unit.workload_state = UnitWorkloadState.CANT_START
+            self.bootstrap_manager.release()
+            return
+
         if not self.cluster_manager.is_healthy:
             event.defer()
             return
--- a/src/charm.py
+++ b/src/charm.py
@@ -100,31 +100,35 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
         )
 
     def _on_bootstrap(self, event: EventBase) -> None:
-        # TODO: logging
-
         if self.state.unit.workload_state != UnitWorkloadState.STARTING:
             if self.bootstrap_manager.try_lock():
+                logger.debug("Bootstrap lock is acquired")
                 if self.workload.is_alive():
+                    logger.debug("Gracefully shutting down an active workload")
                     try:
                         self.node_manager.prepare_shutdown()
                     except ExecError as e:
-                        logger.error(f"Failed to prepare unit for shutdown during restart: {e}")
+                        logger.error(f"Failed to prepare workload shutdown during restart: {e}")
                 self.workload.restart()
                 self.state.unit.workload_state = UnitWorkloadState.STARTING
             event.defer()
             return
 
         if not self._on_bootstrap_pending_check():
-            # TODO: clean restart
-            logger.error("Releasing the bootstrap exclusive lock and ")
+            # TODO: determine whether we need removing var/lib/cassandra/* and in which cases.
+            logger.error(
+                "Releasing the bootstrap exclusive lock and migrating to CANT_START workload state"
+            )
             self.state.unit.workload_state = UnitWorkloadState.CANT_START
             self.bootstrap_manager.release()
             return
 
         if not self.node_manager.is_healthy(ip=self.state.unit.ip):
+            logger.debug("Deferring on_bootstrap due to workload not being healthy yet")
             event.defer()
             return
 
+        logger.debug("Releasing the exclusive lock after successful bootstrap")
         self.bootstrap_manager.release()
 
         if self.state.unit.peer_tls.rotation:
@@ -159,6 +163,8 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
         """Restart Cassandra service."""
         if not self.bootstrap_manager.is_active:
             self.on.bootstrap.emit()
+        else:
+            logger.debug("Restart request was skipped as unit already bootstrapping")
 
     def setup_internal_certificates(self, sans: Sans) -> bool:
         """Configure internal TLS certificates for the current unit using an internally managed CA.
--- a/src/charm.py
+++ b/src/charm.py
@@ -8,11 +8,10 @@ import logging
 
 from charms.data_platform_libs.v1.data_models import TypedCharmBase
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
-from charms.rolling_ops.v0.rollingops import RollingOpsManager, RunWithLock
-from ops import ModelError, SecretNotFoundError, main
-from tenacity import Retrying, stop_after_delay, wait_exponential
+from ops import EventBase, ModelError, SecretNotFoundError, main
 
-from common.exceptions import BadSecretError
+from common.exceptions import BadSecretError, ExecError
+from common.lock_manager import LockManager
 from core.config import CharmConfig
 from core.literals import CASSANDRA_ADMIN_USERNAME
 from core.state import (
@@ -41,6 +40,9 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
     def __init__(self, *args):
         super().__init__(*args)
 
+        self.on.define_event("bootstrap", EventBase)
+        self.framework.observe(self.on.bootstrap, self._on_bootstrap)
+
         self.state = ApplicationState(self)
         self.workload = CassandraWorkload()
         self.cluster_manager = ClusterManager(workload=self.workload)
@@ -62,9 +64,7 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
             user=CASSANDRA_ADMIN_USERNAME,
             password=self.state.cluster.operator_password_secret,
         )
-        bootstrap_manager = RollingOpsManager(
-            charm=self, relation="bootstrap", callback=self.bootstrap
-        )
+        self.bootstrap_manager = LockManager(self, "bootstrap")
 
         self.cassandra_events = CassandraEvents(
             self,
@@ -73,10 +73,10 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
             cluster_manager=self.cluster_manager,
             config_manager=config_manager,
             database_manager=database_manager,
-            bootstrap_manager=bootstrap_manager,
             tls_manager=self.tls_manager,
             setup_internal_certificates=self.setup_internal_certificates,
             read_auth_secret=self.read_auth_secret,
+            restart=self.restart,
         )
 
         self.tls_events = TLSEvents(
@@ -85,9 +85,9 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
             workload=self.workload,
             cluster_manager=self.cluster_manager,
             config_manager=config_manager,
-            bootstrap_manager=bootstrap_manager,
             tls_manager=self.tls_manager,
             setup_internal_certificates=self.setup_internal_certificates,
+            restart=self.restart,
         )
 
         self._grafana_agent = COSAgentProvider(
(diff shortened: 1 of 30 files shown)


src/common/lock_manager.py
```diff
@@ -0,0 +1,239 @@
...
+        self.app = app
+        self.units = [*relation.units, unit]
+
+    def __iter__(self):
+        """Yield a lock for each unit we can find on the relation."""
+        for unit in self.units:
+            yield Lock(relation=self.relation, app=self.app, unit=unit)
+
+
+class LockManager(Object):
```
deusebio, line 166: **todo** this is not a manager, this is an EventHandler it if deals with the implementations of the hooks

src/common/lock_manager.py
```diff
@@ -0,0 +1,239 @@
+#!/usr/bin/env python3
```
deusebio, line 1: **suggestion (non-blocking)** Maybe we could also split the various classes into the "sections":

1. `/src/common/lock/core.py` => domain objects (e.g. `LockState`, `Lock`)
2. `/src/common/lock/events.py` => LockEventHandlers (the manager should be renamed)

src/charm.py
```diff
@@ -41,9 +40,12 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
     def __init__(self, *args):
         super().__init__(*args)
 
+        self.on.define_event("bootstrap", EventBase)
```
deusebio, line 43: **suggestion** maybe we could have a more specific event, like a "BootstrapEvent", or something with a more meaninful name.

tests/integration/test_scaling.py
```diff
@@ -47,7 +46,7 @@ def test_read(juju: jubilant.Juju, app_name: str) -> None:
 
 def test_scale(juju: jubilant.Juju, app_name: str) -> None:
     juju.add_unit(app_name, num_units=2)
-    juju.wait(jubilant.all_active)
+    juju.wait(jubilant.all_active, timeout=1200)
```
deusebio, line 49: **question, suggestion** does the fact that we have increased the timeout mean that with the new business logic it takes longer for the charm to settle? Why does it take so long to settle with two units? 20 minutes seems a bit too much. 

I'm wondering if this is because we wait for update statuses. If that's the case, we can increase the frequency with which update statuses are fired. There used to be one function on pytest-operator (`fast_forward`), I would be surprised if there is not a way to increase that (in the worst case by just resetting the model-config).

deusebio (approved):
This is **TRULY** remarkable work @Zvirovyi ! I have really enjoyed reading through the code, navigating through the various modules. The locks business logic and implementation seems very solid, and shows serious thinking.

The code base is very well structured and commented, and I could follow rather along, despite the number of code lines. Despite the complexity of the workflows, the structure has helped me to understand the flow. 

The only concern that I have here is that I noticed that the time required to be waited in the integration tests seems rather large, for even simply operations (e.g. adding two units). I'm wondering if we can reduce that, because otherwise it may extend any testing of Cassandra, a bit more complex. We can also address this in a separate PR where we try to improve the velocity of the various operations. But overall, if I had to chose, I would always pick more solid processes over velocity. So if this is an artefact of the better stability (and maintainability) provided by the improvements in the bootstrap process, that's fine. We had similar issues in Kafka, and also in that case we often traded velocity for robustness.

src/common/lock_manager.py
Zvirovyi: Okay, I can split the implementation into the Events and Manager:
```python
class LockManager:
    def __init__(self, model: Model, relation_name: str) -> None:
        self.app = model.app
        self.unit = model.unit
        self.relation_name = relation_name
        self.relation = model.get_relation(relation_name)
...

class LockEvents(Object):
    def __init__(self, charm: CharmBase, manager: LockManager):
        super().__init__(charm, key=f"{manager.relation_name}_lock_manager")

        self.manager = manager

        self.framework.observe(
            charm.on[manager.relation_name].relation_changed,
            self.handle_event,
        )
        self.framework.observe(
            charm.on[manager.relation_name].relation_changed,
            self.handle_event,
        )
        self.framework.observe(
            charm.on[manager.relation_name].relation_broken,
            self.handle_event,
        )

    def handle_event(self, _: EventBase) -> None:
        self.manager.process_locks()
```

And it will be the next usage:

```python
self.bootstrap_manager = LockManager(model=self.model, relation_name="bootstrap")
self.bootstrap_events = LockEvents(self, manager=self.bootstrap_manager)
```

Will it fit?

src/common/lock_manager.py
Zvirovyi: I'm against of splitting this module into multiple files. This module is under "common" path that holds non-charm-specific modules that can be shared with other charms. From the developer's point of view, it will introduce redundant difficulty of managing (adopting, updating) 2 files that **provide single feature**.
However I agree that it should be renamed, maybe to something like `lock.py` or `locks.py`?

src/charm.py
Zvirovyi: Will it be okay to put BootstrapEvent class under the CassandraCharm definition? Or will it be better to create `src/core/events.py` and put it here?

tests/integration/test_scaling.py
Zvirovyi: Yes, you're right. While prior, bootstrap hook were immediately called by RollingOpsManager, now we're relying on deferring it. So, I've set `update-status-hook-interval=60s` globally (during testing model creation). Below are some testing values.

Manual deployment of 4 units:
* default value: 17 minutes
* 60s: 9.5 minutes
* 15s: 9 minutes

This PR's integration tests duration: 
* default value: 23272 seconds
* 60s: 21303 seconds (-8.5%)
* default value, reference on main branch: 19523 seconds

src/events/tls.py
```diff
@@ -161,18 +163,27 @@ def _handle_certificate_available_event(
...
         )
+
+        if tls_changed:
+            tls_state.rotation = True
+
+        if not tls_state.rotation:
+            return
+
+        # TODO: logging
+        if not self.state.unit.is_config_change_eligible:
```
welpaolo, line 180: Is there any case in which this condition cannot be reached due to this condition: `not tls_state.rotation`

src/charm.py
```diff
@@ -99,18 +99,37 @@ def __init__(self, *args):
...
         )
 
-    def bootstrap(self, event: RunWithLock) -> None:
-        """Start workload and join this unit to the cluster."""
-        self.state.unit.workload_state = UnitWorkloadState.STARTING
+    def _on_bootstrap(self, event: EventBase) -> None:
+        if self.state.unit.workload_state != UnitWorkloadState.STARTING:
+            if self.bootstrap_manager.try_lock():
+                logger.debug("Bootstrap lock is acquired")
+                if self.workload.is_alive():
```
welpaolo, line 108: Can you put some comments to make more clear the bootstrap logic ?

src/charm.py
```diff
@@ -99,18 +99,37 @@ def __init__(self, *args):
...
+                    logger.debug("Gracefully shutting down an active workload")
+                    try:
+                        self.node_manager.prepare_shutdown()
```
welpaolo, line 111: Do we need to drain every time we restart the unit ?

tests/integration/test_full_cluster_restart.py
```diff
@@ -22,7 +22,7 @@ def test_deploy(juju: jubilant.Juju, cassandra_charm: Path, app_name: str) -> No
     juju.wait(
         ready=lambda status: jubilant.all_agents_idle(status) and jubilant.all_active(status),
         delay=20,
-        timeout=1000,
+        timeout=1800,
```
welpaolo, line 27: Does the new bootstrap process requires a lot more time to bootstrap the cluster ?

welpaolo: I would like to have writes in the database before the restart.

welpaolo (approved):
Really nice work! I left some minor comments

src/charm.py
Zvirovyi: Ideally, yes. Just to have a "safe shutdown" mechanism

src/charm.py
Zvirovyi: What places should be more clarified?

src/events/tls.py
Zvirovyi: When Cassandra restart (or bootstrap) was completed (essentially applying new configuration) before the last execution of this deferred event.

tests/integration/test_full_cluster_restart.py
Zvirovyi: Check [thread above](https://github.com/canonical/cassandra-operator/pull/17#discussion_r2387004182) on bootstrapping duration.

src/common/lock_manager.py
deusebio: Not really, IMHO. If it is a manager, it should not refer to "relation", "unit", "app", etc. One of the idea is that Managers are ops code-free as much as possible. I would have just renamed "LockManager" into "LockEvents". That's it!

src/common/lock_manager.py
deusebio: It was to just provide the same structure `core`,`managers`, and `handlers` using submodules. Yes, the submodules become rather short (but coming from also Java background, I'm not that allergic to that). 

But I'm not super firm on this, it was just a suggestion. Up to you! So if you are not keen, just feel free to resolve the comment.

src/charm.py
deusebio: I was not about the position, more about subclassing a specific event. But for the position, I believe it should be in `src/events` since this pertains to the hook/charm logic. But again, not super firm on this, it is a bit of niptick and indeed a non-blocking suggestion

tests/integration/test_scaling.py
deusebio: yes, if we can speed up the integration tests that would help. It is good to know, and possibly worth outlining in the docs as well, that fast-forwarding update statuses can help. Feel free to resolve the comment

deusebio (approved):
Looks good! All my comments were non-blocking. Thank you for speeding up a bit the integration tests!

src/common/lock_manager.py
Zvirovyi: I agree that manager naming isn't quite a fit for it, but:
1. we don't have and didn't want for any EventHandler to have an API, that will be used in other places. [Our previous **unresolved** discussion](https://github.com/canonical/cassandra-operator/pull/9#discussion_r2208682316) ended on placing shared code into the charm.py file.
2. such way we won't be following the same naming convention of RollingOpsManager.
3. this module is placed under `common` path, that is intended for holding modules, that can be shared with other charms. I see that this module might be useful for some other charms, but in my opinion `LockEvents` will be a bad naming for the _library_, that provides API.

src/charm.py
Zvirovyi: Then we'll probably have a library resolving conflict like that :)
зображення

tests/integration/test_full_cluster_restart.py
Zvirovyi: Also done with writes.

src/charm.py
welpaolo: I will put a comment on why the first condition is needed :) but it is not blocking

src/charm.py
welpaolo: ok perfect

src/common/lock_manager.py
deusebio: > such way we won't be following the same naming convention of RollingOpsManager.

That's true, but the `RollingOpsManager` actually precedes the pattern, and when we did not created the notion of "Manager" and "Handlers". Now that we have this, I'd prefer to use it. But it is not really super critical, we can also address this in the future

src/charm.py
deusebio: non blocking suggestion if it is not straightforward

PR #18, opened 2025-09-23
[DPE-8393] | Scaling down one unit

BON4:
### PR Summary

This PR implements proper handling of node decommissioning when a Juju unit is removed, ensuring safe removal and clear logging.

### Key Changes

* **\_on\_storage\_detaching hook** – Added to handle single-unit removal with proper decommissioning.
* **Health checks** – Cluster health is verified before attempting decommissioning.
* **Logging improvements** – Warnings and errors are logged clearly for easier debugging.
* **Tests** – Comprehensive unit and integration tests covering decommissioning scenarios.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -66,9 +66,11 @@ requires:
     optional: true
 
 storage:
-  cassandra:
+  data:
     type: filesystem
-    location: /var/snap/charmed-cassandra/common
+    location: /var/snap/charmed-cassandra/common/var/lib/cassandra
+    minimum-size: 1G
+    description: storage for cassandra data    
 
 parts:
   # "poetry-deps" part name is a magic constant
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -65,6 +65,11 @@ requires:
     limit: 1
     optional: true
 
+storage:
+  cassandra:
+    type: filesystem
+    location: /var/snap/charmed-cassandra/common
+
 parts:
   # "poetry-deps" part name is a magic constant
   # https://github.com/canonical/craft-parts/pull/901
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -4,6 +4,7 @@
 
 """Handler for main Cassandra charm events."""
 
+from datetime import timedelta
 import logging
 from typing import Callable
 
@@ -20,12 +21,13 @@ from ops import (
     RelationDepartedEvent,
     SecretChangedEvent,
     StartEvent,
+    StorageDetachingEvent,
     UpdateStatusEvent,
 )
 from pydantic import ValidationError
-from tenacity import Retrying, stop_after_delay, wait_exponential, wait_fixed
+from tenacity import Retrying, stop_after_attempt, stop_after_delay, wait_exponential, wait_fixed, stop_never, before_log, after_log, before_sleep_log
 
-from common.exceptions import BadSecretError
+from common.exceptions import BadSecretError, ExecError
 from core.config import CharmConfig
 from core.literals import CASSANDRA_ADMIN_USERNAME
 from core.state import PEER_RELATION, ApplicationState, UnitWorkloadState
@@ -81,6 +83,7 @@ class CassandraEvents(Object):
         self.framework.observe(self.charm.on.update_status, self._on_update_status)
         self.framework.observe(self.charm.on.collect_unit_status, self._on_collect_unit_status)
         self.framework.observe(self.charm.on.collect_app_status, self._on_collect_app_status)
+        self.framework.observe(self.charm.on.cassandra_storage_detaching, self._on_storage_detaching)
 
     def _on_install(self, _: InstallEvent) -> None:
         self.workload.install()
@@ -169,7 +172,6 @@ class CassandraEvents(Object):
             with attempt:
                 self.database_manager.init_admin(password)
         self.state.cluster.operator_password_secret = password
-
         self.cluster_manager.prepare_shutdown()
 
     def _start_subordinate(self, event: StartEvent) -> None:
@@ -283,8 +285,7 @@ class CassandraEvents(Object):
                 )
                 event.defer()
                 return
-            self.cluster_manager.decommission()
-        elif self.charm.unit.is_leader():
+        if self.charm.unit.is_leader():
             self._recover_seeds(event)
 
     def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
@@ -403,3 +404,19 @@ class CassandraEvents(Object):
             and bool(old_hostname)
             and (old_ip != self.state.unit.ip or old_hostname != self.state.unit.hostname)
         )
+
+    def _on_storage_detaching(self, _: StorageDetachingEvent) -> None:
+        if not self.cluster_manager.cluster_healthy():
+            raise Exception("Cluster is not healthy, cannot remove unit")
+
+        if self.charm.app.planned_units() < len(self.state.units) - 1:
+            logger.warning(f"More than one unit removing at a time is not supported. The charm may be in a broken, unrecoverable state")
+        
+        logger.info(f"Starting unit {self.state.unit.unit_name} node decommissioning")
+        try:
+            self.cluster_manager.decommission()
+        except ExecError as e:
+            logger.error(f"Failed to decommission unit: {e}")
+            raise e
+        
+        logger.info("Storage deatached")
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -34,7 +34,7 @@ from tenacity import (
 from common.exceptions import BadSecretError, ExecError
 from core.config import CharmConfig
 from core.literals import CASSANDRA_ADMIN_USERNAME
-from core.state import PEER_RELATION, ApplicationState, UnitWorkloadState
+from core.state import PEER_RELATION, ApplicationState, UnitWorkloadState, DATA_STORAGE
 from core.statuses import Status
 from core.workload import WorkloadBase
 from managers.cluster import ClusterManager
@@ -88,7 +88,7 @@ class CassandraEvents(Object):
         self.framework.observe(self.charm.on.collect_unit_status, self._on_collect_unit_status)
         self.framework.observe(self.charm.on.collect_app_status, self._on_collect_app_status)
         self.framework.observe(
-            self.charm.on.cassandra_storage_detaching, self._on_storage_detaching
+            self.charm.on[DATA_STORAGE].storage_detaching, self._on_storage_detaching
         )
 
     def _on_install(self, _: InstallEvent) -> None:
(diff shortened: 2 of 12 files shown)


src/managers/cluster.py
```diff
@@ -24,14 +34,21 @@ def __init__(self, workload: WorkloadBase):
...
+
         try:
             stdout, _ = self._workload.exec([_NODETOOL, "info"], suppress_error_log=True)
             return "Native Transport active: true" in stdout
         except ExecError:
             return False
 
     def network_address(self) -> tuple[str, str]:
-        """Get hostname and IP of this unit."""
+        """Get IP and hostname of this unit."""
```
Zvirovyi: Nice catch!

tests/unit/test_scale.py
```diff
@@ -0,0 +1,107 @@
...
+            "managers.cluster.ClusterManager.decommission",
+            side_effect=ExecError(stdout="", stderr="error"),
+        ),
+    ):
+        units.return_value = [MagicMock()]  # один юнит
+        # Проверяем, что ctx.run вызывает UncaughtCharmError
+        with pytest.raises(scenario.errors.UncaughtCharmError) as e:
+            ctx.run(ctx.on.storage_detaching(storage), state)
+
+        # Внутри e.value.__cause__ хранится оригинальный ExecError
```
Zvirovyi: Let's maintain codebase in English.

tests/unit/test_scale.py
Zvirovyi, line 1: These unit tests are failing on MacOS:
```
FAILED tests/unit/test_scale.py::test_storage_detaching_cluster_unhealthy - AssertionError: Regex pattern did not match.
FAILED tests/unit/test_scale.py::test_storage_detaching_multiple_units_removal_logs_warning - scenario.errors.UncaughtCharmError: Uncaught SnapError in charm, try "exceptions [n]"
 if using pdb on Python 3.13+. Details: <charms.operator_libs_linux.v2.snap.SnapError ('s...
FAILED tests/unit/test_scale.py::test_storage_detaching_success - scenario.errors.UncaughtCharmError: Uncaught SnapError in charm, try "exceptions [n]" if using pdb on Python 3.13
+. Details: <charms.operator_libs_linux.v2.snap.SnapError ('s...
FAILED tests/unit/test_scale.py::test_storage_detaching_decommission_fails - assert False
```

Full log: [tmp.log](https://github.com/user-attachments/files/22492821/tmp.log)

Zvirovyi (commented):
Good work overall! Have a couple of questions though

charmcraft.yaml
```diff
@@ -65,6 +65,11 @@ requires:
     limit: 1
     optional: true
 
+storage:
+  cassandra:
+    type: filesystem
+    location: /var/snap/charmed-cassandra/common
```
deusebio: **todo** could we set this to 

```
/var/snap/charmed-cassandra/common/var/lib/cassandra
```

We could also add the minimum requirements, a bit like in [here](https://github.com/canonical/kafka-operator/blob/main/metadata.yaml#L71).

src/events/cassandra.py
```diff
@@ -403,3 +410,22 @@ def _update_network_address(self) -> bool:
...
+            )
+
+        logger.info(f"Starting unit {self.state.unit.unit_name} node decommissioning")
+        try:
+            self.cluster_manager.decommission()
+        except ExecError as e:
+            logger.error(f"Failed to decommission unit: {e}")
+            raise e
+
+        logger.info("Storage deatached")
```
deusebio: **niptick** I believe there is another event which is called "Storage detached". Maybe it is better to log something like:

```
storage-detaching event hook completed
```

deusebio (commented):
Overall it looks good. 

I actually didn't realize we were not having external storage. Good that we have provided support for this. 

I just have a small comment/question: could we use multiple storages or Cassandra would always just work with one? I'm taking inspiration from Kafka where we handle multiple storages to have larger IOPs

Zvirovyi (approved):
LGTM

charmcraft.yaml
BON4: Not sure about the number of storage instances to be requested, but `minimum-size`, `description` specified, and `location` is changed.

src/events/cassandra.py
```diff
@@ -20,15 +20,21 @@
...
     UpdateStatusEvent,
 )
 from pydantic import ValidationError
-from tenacity import Retrying, stop_after_delay, wait_exponential, wait_fixed
+from tenacity import (
+    Retrying,
+    stop_after_delay,
+    wait_exponential,
+    wait_fixed,
+)
```
deusebio, line 30: **question** why we are not using this? Was this supported to be used on some retry logic in the hook?

BON4: I don't really know what will be better logic for charm. To retry decommissioning, or to raise after one failed attempt. We could use retrying in future to use locks on multi-unit scaling down.

BON4: @marceloneppel

deusebio:
@BON4 could you fix unit and liting to make sure that the tests work fine?

src/events/cassandra.py
deusebio: let's hear @marceloneppel thoughts. If we don't put the retry logic, are we sure that the next hooks are not running?

BON4: @marcoppenheimer any thoughts too?

marceloneppel: On PostgreSQL, we used to have some retry logic (and if the charm couldn't finish it after the last retry, we just logged that and let the next event run, because what we needed to do at this moment was only to perform a switchover, which the workload would do later if we can't).

As the decommission process of Cassandra seems to be a different case, where the data needs to be moved to another node, I think it's better to raise an exception like the OpenSearch charm does, to avoid that the unit is removed, so the human operator can call `juju resolve-unit` later to retry the decomission process again after fixing what's causing it not to complete.

deusebio: All right! Understood! I believe this is solid reasoning! Thank you for chiming in!

tests/integration/helpers.py
```diff
@@ -302,3 +302,89 @@ def all_prometheus_exporters_data(juju: jubilant.Juju, check_field: str, app_nam
     for unit in status.apps[app_name].units.values():
         result = result and check_field in (prometheus_exporter_data(unit.public_address) or "")
     return result
+
+
+def prepare_keyspace_and_table(
+    juju: jubilant.Juju, app_name: str, ks="test", table="kv", unit_name: str = ""
```
theoctober19th: nit: type `ks` and `table` as `str`

tests/integration/helpers.py
```diff
@@ -302,3 +302,89 @@ def all_prometheus_exporters_data(juju: jubilant.Juju, check_field: str, app_nam
...
+    if unit_name:
+        if unit_name not in units:
+            raise ValueError(f"Unit {unit_name} not found in app {app_name}")
+        hosts = [units[unit_name].public_address] + [
+            u.public_address for name, u in units.items() if name != unit_name
+        ]
+    else:
+        hosts = [u.public_address for u in units.values()]
+
+    hosts = [u.public_address for u in juju.status().apps[app_name].units.values()]
```
theoctober19th: I see that line `321` is just repeating what has been done in earlier lines, and the list is being overwritten.

tests/integration/helpers.py
```diff
@@ -302,3 +302,89 @@ def all_prometheus_exporters_data(juju: jubilant.Juju, check_field: str, app_nam
...
+        if unit_name not in units:
+            raise ValueError(f"Unit {unit_name} not found in app {app_name}")
+        hosts = [units[unit_name].public_address] + [
+            u.public_address for name, u in units.items() if name != unit_name
+        ]
+    else:
+        hosts = [u.public_address for u in units.values()]
+
+    got = {}
+    with connect_cql(
```
theoctober19th, line 363: suggestion: I see that a the calculation of `hosts` and the initialization of the session is being repeated multiple times with the only difference being the queries and interpretation of the results -- would it be possible to reuse this logic?

tests/integration/test_scaling.py
```diff
@@ -22,49 +22,62 @@ def test_deploy(juju: jubilant.Juju, cassandra_charm: Path, app_name: str) -> No
...
-        res = session.execute("SELECT message FROM test", timeout=300)
-        assert (
-            isinstance(res, ResultSet)
-            and len(res_list := res.all()) == 1
-            and res_list[0].message == "hello"
-        ), "test data written prior aren't found"
+    juju.add_unit(app_name, num_units=2)
+    juju.wait(jubilant.all_active)
 
+    added_units = [u[0] for u in juju.status().apps[app_name].units.items() if not u[1].leader]
```
theoctober19th: question: I'm not able to understand how `u[0]` is the newly added unit -- since with this logic `u[0]` will be the leader unit, which will probably be the one that has been there since before. Maybe you meant to have `u[1]`? Or maybe store the existing units somewhere before, and calculate what was the one that was added?

tests/integration/test_scaling.py
```diff
@@ -22,49 +22,62 @@ def test_deploy(juju: jubilant.Juju, cassandra_charm: Path, app_name: str) -> No
...
+    juju.remove_unit(non_leader_units[0])
     juju.wait(jubilant.all_active)
 
+    got = read_n_rows(juju, app_name, ks=ks, table=tb, unit_name=non_leader_units[1])
+    assert_rows(wrote, got)
+
+    juju.remove_unit(leader_unit)
+    juju.wait(jubilant.all_active)
+
+    new_leader_unit = [
```
theoctober19th, line 88: praise: nice to have the scenario of removing the leader unit as well.

theoctober19th (commented):
I've left a few comments to better understand the context behind some changes.

tests/integration/test_scaling.py
deusebio: @BON4 Bikalpa is OoO for a few days. Please just provide an answer to the question above but then let's feel free to proceed the PR towards merging the contribution

deusebio (approved):
Looks good! 

There are some conflicts on the `charmcraft.yaml` file, but even more importatly I can probably foresee conflicts with #17 too, which is also about to be merged. So let's sync with @Zvirovyi on who merges first, to be mindful of each other work.

If anything, I would honestly first merge #17 which has a broader scope and for which rebasing may be more tricky.

tests/integration/helpers.py
BON4: Thanks! Have not noticed this.

tests/integration/test_scaling.py
BON4: So, in the start we had 1 unit, then we scale up to 3. So, all non-leader units is new units. I fixed it, so it will be more readable.

welpaolo (approved):
LGTM

PR #20, opened 2025-09-24
[DPE-8410] Enable TLS for DatabaseManager

Zvirovyi:
Currently, database manager doesn’t account of TLS client relation. So, if that relation is established, password change & seed nodes pinging (introduced in handling full cluster restart PR) operations will fail. This PR fixes it.

### DIFF
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -53,12 +53,24 @@ class CassandraPaths:
         return self.data_dir / "saved_caches"
 
     def get_truststore(self, scope: TLSScope) -> pathops.PathProtocol:
-        """Get keystore path for the scope."""
+        """Get truststore path for the TLS scope."""
         return self.tls_dir / f"{scope.value}-truststore.jks"
 
     def get_keystore(self, scope: TLSScope) -> pathops.PathProtocol:
-        """Get keystore path for the scope."""
+        """Get keystore path for the TLS scope."""
         return self.tls_dir / f"{scope.value}-keystore.p12"
+
+    def get_ca(self, scope: TLSScope) -> pathops.PathProtocol:
+        """GET ca path for the TLS scope."""
+        return self.tls_dir / f"{scope.value}-ca.pem"
+
+    def get_certificate(self, scope: TLSScope) -> pathops.PathProtocol:
+        """Get certificate path for the TLS scope."""
+        return self.tls_dir / f"{scope.value}-unit.pem"
+
+    def get_private_key(self, scope: TLSScope) -> pathops.PathProtocol:
+        """Get private key path for the TLS scope."""
+        return self.tls_dir / f"{scope.value}-private.key"
 
     @property
     def jmx_exporter(self) -> pathops.PathProtocol:
--- a/src/core/workload.py
+++ b/src/core/workload.py
@@ -61,7 +61,7 @@ class CassandraPaths:
         return self.tls_dir / f"{scope.value}-keystore.p12"
 
     def get_ca(self, scope: TLSScope) -> pathops.PathProtocol:
-        """GET ca path for the TLS scope."""
+        """Get ca path for the TLS scope."""
         return self.tls_dir / f"{scope.value}-ca.pem"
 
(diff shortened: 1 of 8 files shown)


src/managers/database.py
```diff
@@ -29,18 +32,37 @@
...
+        password: str,
+        enable_ssl: bool = False,
+    ):
         self.execution_profile = ExecutionProfile(
             load_balancing_policy=TokenAwarePolicy(DCAwareRoundRobinPolicy())
         )
         self.auth_provider = (
             PlainTextAuthProvider(username=user, password=password) if user and password else None
         )
+        if enable_ssl:
```
deusebio: **suggestion(non-blocking)** we could also allow the manager to auto-infer this, no? if the tls files exists, it means that we have tls enabled. This way we would also avoid to make mistakes (like enabling ssl when there are no files there or not using tls when the server is configured in that way)

deusebio (approved):
Looks good! Just a small non-blocking suggestion

src/managers/database.py
Zvirovyi: @BON4 what do you think? Is it better to rely on `state.unit.client_tls.ready` or on files existence?

BON4: I think relying on the existence of files is a better approach.
The certificates are first propagated through relation data, but ultimately they end up as files on disk. Cassandra itself depends on those files, not on the relation data directly.

Also we already have `client_tls_ready` in `tls_manager`, and it is checks existence of a client keystore and truststore, which are created last. We can use this function as it is, or modify it to include other certs.

src/core/workload.py
```diff
@@ -53,13 +53,25 @@ def saved_caches_directory(self) -> pathops.PathProtocol:
...
+        """Get truststore path for the TLS scope."""
         return self.tls_dir / f"{scope.value}-truststore.jks"
 
     def get_keystore(self, scope: TLSScope) -> pathops.PathProtocol:
-        """Get keystore path for the scope."""
+        """Get keystore path for the TLS scope."""
         return self.tls_dir / f"{scope.value}-keystore.p12"
 
+    def get_ca(self, scope: TLSScope) -> pathops.PathProtocol:
+        """GET ca path for the TLS scope."""
```
welpaolo:
```suggestion
        """Get ca path for the TLS scope."""
```

welpaolo (approved):
LGTM.

deusebio:
@Zvirovyi are you planning to address the comment (non-blocking) above?

src/managers/database.py
Zvirovyi: So, let's then continue [our previous thread on bringing some business logic into workload](https://github.com/canonical/cassandra-operator/pull/9#discussion_r2207686259). I don't want to use any methods from single manager in the another one (they should be fully separated). We can bring the next method into CassandraPaths, that will allow us to reuse same logic in multiple places (including TLS manager itself), although we agreed to not do this before:

```python
def is_tls_configured(self, scope: TLSScope) -> bool:
    return self.get_truststore(scope).exists() and self.get_keystore(scope).exists()
```

My question is, if it's only the proxy for the workload methods itself, would be it a fit for the CassandraPaths inside Workload?

deusebio: Having a manager inside another manager does not seem wrong, there may be more more high-level managers and more low-level (product specific) managers. For instance, imagine there is a general K8s manager with the business logic to check validity of service account, namespace compliance, etc, that is used on a more Cassandra specific manager.

Path in general only specifies the paths, don't perform any check, for which they would require access to the container as well, which the Path does not have, I believe. 

But let's not overstress about this. My ask was to just have a auto-inference logic to prevent mistake. I don't want to be too prescriptive where. The manager would feel a proper place, but again, use your best judgement without overcomplicating this.

Zvirovyi: @marcoppenheimer what do you think?

Zvirovyi: We agreed on using TLS manager inside Database one.

deusebio (approved):
Looks good!

PR #21, opened 2025-10-07
[DPE-8323] Implement client relation

BON4:
### **Description**

This PR introduces support for **external client relations** in the Cassandra Operator using the new experimental [`charms.data_platform_libs.v1.data_interfaces`](https://github.com/canonical/data-platform-libs/pull/241) library.

---

#### **Provider side**

* Implemented new `ExternalClientsEvents` handler.
* Added support for:

  * **Keyspace requests** (`_on_resource_requested`)
  * **Entity (user) requests** (`_on_resource_entity_requested`)
  * **Permission updates** (`_on_resource_entity_permissions_changed`)
  * **Cleanup on relation broken** (`_on_relation_broken`)

#### **Requirer side (test charm)**

* Added integration charm `tests/integration/application-charm` that acts as a **client**.
* Uses `ResourceRequirerEventHandler` to:

  * Request keyspace and user resources.
  * Update user permissions dynamically via config.
  * Handle endpoint changes and created events.
* Provides a working end-to-end example of the new data contract flow.
* Includes CQL connectivity tests via `cqlsh` snap and Cassandra Python driver.
* Includes CQL tests with client TLS enabled.  

---

### **Testing**

Integration tested with the following scenarios:

1. Client requests a keyspace — provider creates keyspace and returns credentials for default user.
2. Client requests a user entity with specific permissions — provider grants them.
3. Config change on requirer updates user permissions — provider updates Cassandra roles.
4. Client removes relation — provider deletes corresponding roles.
5. End-to-end test with table creation through the requirer charm using returned credentials.

---

### **Known limitations**

* `mtls_enabled=False` (mTLS client support is not yet implemented; placeholder only).
* `EntityPermissionModel` currently ignores `resource_type`, assuming keyspace-level scope only.
* `_on_resource_requested` - also creates a default user. Logically it should create resource only. There is no way to respond to this event without specifying username and password in response. 
* `data_interfaces` v1 library is still experimental and may introduce breaking changes.

---

### **Bug fixes**

* `self-signed-certificates` is updated to `1/stable` with ubuntu@24.04 base.
* `client_encryption_options.require_client_auth` value is chaged to False, becouse it is used for mTLS only.


tests/integration/application-charm/.gitignore
Mehdi-Bendriss, line 1: nit: small typo in the name of the directory `application-charm`

src/charm.py
```diff
@@ -101,7 +118,7 @@ def __init__(self, *args):
             log_slots=[f"{SNAP_NAME}:logs"],
         )
 
-    def _on_bootstrap(self, event: EventBase) -> None:
+    def _on_bootstrap(self, event: EventBase) -> None:  # noqa: C901
```
Zvirovyi: **suggestion** I think we can avoid ignoring C901 rule by moving `if self.state.unit.workload_state != UnitWorkloadState.STARTING:` block into distinct function, just like with `_on_bootstrap_pending_check`.

src/charm.py
```diff
@@ -92,6 +99,16 @@ def __init__(self, *args):
...
         )
 
+        self.provider_events = ExternalClientsEvents(
+            self,
+            state=self.state,
+            workload=self.workload,
+            node_manager=self.node_manager,
+            tls_manager=self.tls_manager,
+            database_manager=database_manager,
+            acquire_operator_password=self.acquire_operator_password,
```
Zvirovyi: **todo** there's no need to read password that way in ExternalClientsEvents. For any external client event state password (that's already populated in DatabaseManager) should be valid because password change handling only happens in CassandraEvents, and state password is always pointing to the live Cassandra password.

src/charm.py
```diff
@@ -216,6 +238,26 @@ def setup_internal_certificates(self, sans: Sans) -> bool:
...
+
+        Raises:
+            BadSecretError: If the configured Juju secret cannot be found or is invalid.
+        """
+        if self.config.system_users:
+            return self.read_auth_secret(self.config.system_users)
+        if self.state.cluster.operator_password_secret:
+            return self.state.cluster.operator_password_secret
+        return self.workload.generate_string()
+
```
Zvirovyi: **todo** this function is redundant as password handling happens only in CassandraEvents.

src/managers/tls.py
```diff
@@ -483,6 +492,36 @@ def remove_stores(self, scope: TLSScope = TLSScope.CLIENT) -> None:
...
+                continue
+
+            self.update_cert(entity.alias, entity.cert.raw, trust_password, scope)
+            live_aliases.add(entity.alias)
+            should_reload = True
+
+        if should_reload:
+            logger.debug(f"Following aliases should be in the {scope} truststore: {live_aliases}")
+            self.reload_truststore(scope)
+
```
Zvirovyi: **question** are these functions used anywhere? I don't see any references

src/managers/database.py
```diff
@@ -112,15 +163,107 @@ def init_admin(self, password: str) -> None:
         ) as session:
             session.execute("DROP ROLE %s", [_CASSANDRA_DEFAULT_CREDENTIALS])
 
-    def update_role_password(self, user: str, password: str) -> None:
+    def create_keyspace(self, ks: str, rf: int, op_password: str) -> None:
```
Zvirovyi: **todo** I think we can remove `op_password` as it should always be live in the state password from constructor. Same applies to other functions.

src/managers/database.py
```diff
@@ -30,6 +31,56 @@
...
+
+    _valid_permissions = {
+        "ALTER",
+        "AUTHORIZE",
+        "DESCRIBE",
+        "DROP",
+        "MODIFY",
+        "SELECT",
+        "CREATE",
+    }
```
Zvirovyi, line 48: **suggestion** let's move this to the static method. This way we'll not be in need of `get_all_valid_permissions` proxy function.

src/events/cassandra.py
Zvirovyi, line 1: **todo** along with the comments from `charm.py`, these changes should be reverted.

src/events/provider.py
Zvirovyi, line 1: **suggestion** let's either rename this file to `external_clients.py` or class to `ProviderEvents`. Just to keep consistency between class and file names. Same with `test_provider.py` integration & unit tests.

src/events/provider.py
```diff
@@ -0,0 +1,320 @@
...
+
+        if not self.charm.unit.is_leader():
+            return
+
+        if any(
+            [
+                self.state.unit.workload_state != UnitWorkloadState.ACTIVE,
+                not self.workload.is_alive(),
+            ]
+        ):
```
Zvirovyi: **suggestion** it should be safe to rely only on `if self.state.unit.is_operational`. We already do checks this way in other places. Also applies to other funcs.

src/events/provider.py
```diff
@@ -0,0 +1,320 @@
...
+            logging.debug("Deferring _on_resource_requested waiting for MTLS setup")
+            event.defer()
+            return
+
+        rolename = self._rolename_from_relation(relation_id, request.salt)
+        password = self.workload.generate_string()
+
+        if rolename in self.state.cluster.roles:
+            logger.error(f"Rolename: {rolename} already exists for this relation")
```
Zvirovyi, line 104: **question** @deusebio @marcoppenheimer should we do more than error log?

src/events/provider.py
```diff
@@ -0,0 +1,320 @@
...
+            logger.debug("Deferring _on_relation_broken inactive unit workload state")
+            event.defer()
+            return
+
+        # don't remove anything if app is going down
+        if self.charm.app.planned_units() == 0:
+            return
+
+        if event.relation.app != self.charm.app:
+            for rolename in self.state.cluster.roles:
```
Zvirovyi: **todo** there's a critical bug. Let's assume we have 3 clients. Then, relation with one of them breaks and **all the roles are deleted** (e.g. 2 other clients have their roles removed here). Tested practically.

tests/integration/helpers.py
```diff
@@ -120,13 +179,20 @@ def check_tls(ip: str, port: int) -> bool:
...
     status = juju.status()
     address = status.apps[app_name].units[f"{app_name}/{unit_num}"].public_address
     return address
 
 
+def get_unit_names(juju: jubilant.Juju, app_name: str) -> list[str]:
+    """Get the address for a unit."""
+    units = juju.status().apps[app_name].units.items()
+    addresses = [u[0] for u in units]
+    return addresses
```
Zvirovyi, line 193: **todo** wrong description.

Zvirovyi (changes requested):
Have some comments, but truly impressive work!
Haven't checked test_provider integration & unit test, will do this afterward.

src/managers/tls.py
BON4: It was left before the discussion about disabling mTLS. I thought it would be useful in the future. Should I remove them?

src/events/provider.py
BON4: Good notice! I fixed it by moving `roles` to a dataclass with `relation_id` as a field. And unit test added specifically for this issue.

I also noticed that in Kafka this is handled through **relation data** with `relation_interface`. For reference:
* [models.py](https://github.com/canonical/kafka-operator/blob/1cbe13f8abc44938bed1aa0e86c3765cbcfad1f8/src/core/models.py#L874)
* [cluster.py](https://github.com/canonical/kafka-operator/blob/1cbe13f8abc44938bed1aa0e86c3765cbcfad1f8/src/core/cluster.py#L256)

I thought about doing the same, but it's hard to integrate with the existing logic we have in **v0 of data_interfaces** using **v1**. I could spend more time on this.

Doing it this way would make it easier to handle client data, including **mTLS**, in the future.

Any ideas, @deusebio @marcoppenheimer?

src/core/state.py
```diff
@@ -426,6 +432,24 @@ def is_operational(self) -> bool:
...
 
 
+@dataclass(frozen=True)
+class DbRole:
+    """Represents a Cassandra database role linked to a specific relation."""
+
+    name: str
+    relation_id: int
+
+    def to_str(self) -> str:
```
deusebio: **suggestion** what about using `__str__` and then just using the pythoninc version 

```
str(db_role)
```

instead of 

```
db_role.to_str()
```

src/core/workload.py
```diff
@@ -140,10 +140,12 @@ def exec(
         pass
 
     @staticmethod
-    def generate_password() -> str:
+    def generate_string(length: int = 32, salt: str = "") -> str:
```
deusebio: **niptick** 
```suggestion
    def generate_string(length: int = 32, prefix: str = "") -> str:
```

src/managers/tls.py
deusebio: Yes, I'd rather remove these if they are not used

tests/integration/application-charm/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
deusebio: **todo** The LICENSE is already present in the root folder. Please delete this one

tests/integration/application-charm/CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
+# Contributing
```
deusebio: **todo** The contributing section is already present in the root folder. Please delete this one

tests/integration/application-charm/README.md
```diff
@@ -0,0 +1,26 @@
...
+* documentation > published on (or linked to from) Charmhub
+* detailed contribution guide > documentation or CONTRIBUTING.md
+
+Use links instead.
+-->
+
+# appliation-charm
+
+Charmhub package name: operator-template
+More information: https://charmhub.io/appliation-charm
```
deusebio: **todo** I believe this does not exist. I would remove the README.md unless we convey relevant information. But for a test charm it is not needed.

tests/integration/application-charm/pyproject.toml
```diff
@@ -0,0 +1,88 @@
...
+[tool.poetry.dependencies]
+python = "^3.10"
+ops = ">=2.19.0"
+requests = "^2.32.3"
+requests-unixsocket = "^0.4.1"
+pydantic = "^2.11.4"
+cassandra-driver = "^3.29.2"
+tenacity = "^9.1.2"
+charmlibs-pathops = "^1.0.1"
+cosl = ">= 0.0.50"
```
deusebio: **question** are you sure we need all these packages? some of them are possibly not used in the charm.py of the test application (e.g. `cosl`, `tenacity`). Could you just keep what is strictly necessary?

tests/integration/application-charm/requirements.txt
```diff
@@ -0,0 +1 @@
+ops ~= 2.17
```
deusebio: **todo** I don't think this is needed

tests/integration/application-charm/src/charm.py
```diff
@@ -0,0 +1,351 @@
...
+    ResourceRequirerEventHandler,
+)
+from charms.operator_libs_linux.v2 import snap
+from ops import ActionEvent, InstallEvent
+from ops.charm import CharmBase, ConfigChangedEvent
+from ops.framework import StoredState
+from ops.main import main
+from ops.model import ActiveStatus, WaitingStatus
+
+CQLSH_SNAP_NAME = "cqlsh"
```
deusebio, line 41: **thought** interesting that there exists this snap. I'm wondering whether given the project, DP should take this over, and make sure that the server and client snaps are aligned for the versions

FYI @marcoppenheimer

tests/integration/conftest.py
```diff
@@ -107,9 +107,9 @@ def charm_versions() -> IntegrationTestsCharms:
     return IntegrationTestsCharms(
         tls=TestCharm(
             name="self-signed-certificates",
-            channel="latest/stable",
-            revision=163,  # FIXME (certs): Unpin the revision once the charm is fixed
-            base="ubuntu@22.04",
+            channel="1/stable",
+            revision=317,
+            base="ubuntu@24.04",
```
deusebio, line 112: **praise** well done on pinning proper channel and revision! Great job!

tests/integration/test_provider.py
```diff
@@ -0,0 +1,248 @@
...
+
+    with_new_users = get_db_users(juju, app_name)
+
+    new_users = with_new_users - old_users
+
+    # Assert that the admin user and keyspace were created after a successful integration.
+    # The application charm makes two requests:
+    # 1. For the resource itself (initial user and keyspace)
+    # 2. For the resource entity (regular user)
+    assert len(new_users) == 2
```
deusebio, line 76: **question** I'm not sure I fully follow why we would need to have two users here. Is this consistent with other database logic? @Mehdi-Bendriss

tests/integration/test_provider.py
```diff
@@ -0,0 +1,248 @@
...
+
+    ca = get_cluster_client_ca(juju, app_name)
+
+    requirer_unit = next(iter(juju.status().apps[requirer_app_name].units))
+
+    juju.run(requirer_unit, "create-table", {"table-name": TABLE_NAME + table_prefix})
+
+    juju.wait(jubilant.all_active)
+
+    assert table_exists(juju, app_name, KEYSPACE_NAME, TABLE_NAME + table_prefix, client_ca=ca)
```
deusebio, line 129: **question** just for the sake of it, can we also test that the command does not work if we don't provide a CA or the "right" CA?

tests/integration/test_provider.py
```diff
@@ -0,0 +1,248 @@
+#!/usr/bin/env python3
```
deusebio, line 1: **praise** very well thought test scenario, featuring certs/key rotations, changes in user-permissions, enabling/disabling TLS and user removals! Great job!

src/events/provider.py
```diff
@@ -0,0 +1,312 @@
...
+        if not event.relation or not event.relation.app:
+            logger.debug("Deferring _on_resource_requested event relation is not ready")
+            event.defer()
+            return
+
+        if request.mtls_cert and self.tls_manager.alias_needs_update(
+            self._client_alias_from_relation(self.charm.app.name, relation_id),
+            request.mtls_cert.get_secret_value(),
+        ):
+            logging.debug("Deferring _on_resource_requested waiting for MTLS setup")
```
deusebio: **question** is this relevant given we don't support mTLS?

src/events/provider.py
deusebio: I believe there is a design to report back to the requirer some errors when creating the users, see [DA-174](https://docs.google.com/document/d/1YIRE2XP_xILoec6hOPOPdsiIJPcFqzI7JA-mEw1XEw4/edit?tab=t.0). But I'm not sure this is supported yet. Maybe worth checking either with Iman or Neha

tests/integration/application-charm/src/charm.py
```diff
@@ -0,0 +1,351 @@
...
+        self.cassandra_client = ResourceRequirerEventHandler(
+            self,
+            "cassandra-client",
+            requests=[
+                RequirerCommonModel(resource=self.keyspace_name),
+                RequirerCommonModel(
+                    resource=self.keyspace_name,
+                    entity_type="USER",
+                    entity_permissions=[self._storage_data.keyspace],
+                ),
```
deusebio, line 86: **question** why do we need both? Maybe, just to have a bit more granular test, we could have one endpoint for creating keyspace, and one endpoint to request a user. This would be to possibly test the keyspace creation and lifecycle management independently to the one for users.

src/managers/database.py
```diff
@@ -30,6 +31,57 @@
 _CASSANDRA_DEFAULT_CREDENTIALS = "cassandra"
 
 
+ALL_PERMISSION = "ALL"
+
+
+class Permissions:
```
deusebio, line 37: **praise** Nice touch for handling permissions!

src/events/provider.py
deusebio: Oh yes, this does indeed seem wrong to remove all users. Great catch @Zvirovyi !

Given the logic that we have used in the various hooks, couldn't we just filter for the `rolename` that includes the `relation_id`? I would suppose we would still have access at least to the ID and we could filter the roles based on the `relation_<id>` pattern. However, I could see that this may not be 100% safe. Maybe one way to do this could also be to move this logic from broken event to departed, such that we can still access the entire databag

src/managers/database.py
```diff
@@ -112,15 +163,107 @@ def init_admin(self, password: str) -> None:
...
+
+        with self._session(
+            hosts=self.hosts,
+            auth_provider=PlainTextAuthProvider(
+                username=CASSANDRA_ADMIN_USERNAME, password=op_password
+            ),
+        ) as session:
+            session.execute(cql, [rf])
+
+    def remove_keyspace(self, ks: str, op_password: str) -> None:
```
deusebio: **praise** very clean API for interacting with Keyspaces, Users and Permissions

deusebio (commented):
Very good job @BON4 ! The code is indeed very well structured and easy to follow. I have appreciated many things in this PR, among which:
1. Very good test scenarios
2. Manager functionalities to handle user/keyspace creation and setting permissions
3. Usage of dataclasses (e.g. to represent Permissions)

Apart from a few minor comments, I believe there are only two major points to be addressed before merging are:
1. Fix the bug spotted by @Zvirovyi (I have provided a suggestion, although you are free to achieve the same mean in a different way)
2. I believe it would be good to decouple the keyspace lifecycle from the user lifecycle in the test cham (unless we have strong reasons why they should be attached). This way we could also test that we only remove the user, and we make sure they are roles are still there (to make sure that the bug above is effectively fixed)

tests/integration/application-charm/src/charm.py
BON4: > This would possibly be to test keyspace creation and lifecycle management independently from the one for users.

As for decoupling endpoints - yes, it’s possible to do it this way:

```python
self.cassandra_resource = ResourceRequirerEventHandler(
    self,
    "cassandra-client",
    requests=[
        RequirerCommonModel(resource=self.keyspace_name),
    ],
    response_model=ResourceProviderModel,
)

self.cassandra_entity = ResourceRequirerEventHandler(
    self,
    "cassandra-client",
    requests=[
        RequirerCommonModel(
            resource=self.keyspace_name,
            entity_type="USER",
            entity_permissions=[self._storage_data.keyspace],
        ),
    ],
    response_model=ResourceProviderModel,
)
```

However, there’s an issue when handling `resource_created` events specifically for keyspace creation.
According to the logic in the new `data_interfaces`, the provider **cannot** respond to a `resource_created` event without specifying both `username` and `password`.
If these are missing, the requirer will ignore the response.
You can see this behavior in the code here:
[https://github.com/canonical/data-platform-libs/blob/f92c04b4b8b1e8bb27cde02350abb3e8ad2f9758/lib/charms/data_platform_libs/v1/data_interfaces.py#L2768-L2774](https://github.com/canonical/data-platform-libs/blob/f92c04b4b8b1e8bb27cde02350abb3e8ad2f9758/lib/charms/data_platform_libs/v1/data_interfaces.py#L2768-L2774)

Based on my testing, there are only two ways to achieve the desired behavior:

1. Respond with the default `operator` username and password.
2. Respond with placeholder values, for example:

   ```python
   {"username": "_", "password": "_"}
   ```

src/events/provider.py
BON4: I could not find any built-in functionality for handling errors; it might still be under development.
However, we can extend the base response model, `ResourceProviderModel`, and add custom fields such as `error` or `status`.
The downside is that both the provider and requirer charms would need access to the same extended model — either through a shared library or a synchronized implementation.

tests/integration/test_provider.py
BON4: > However, there’s an issue when handling `resource_created` events specifically for keyspace creation. According to the logic in the new `data_interfaces`, the provider **cannot** respond to a `resource_created` event without specifying both `username` and `password`. If these are missing, the requirer will ignore the response. You can see this behavior in the code here: https://github.com/canonical/data-platform-libs/blob/f92c04b4b8b1e8bb27cde02350abb3e8ad2f9758/lib/charms/data_platform_libs/v1/data_interfaces.py#L2768-L2774

tests/integration/application-charm/src/charm.py
BON4: @welpaolo @marcoppenheimer

src/events/provider.py
deusebio: Please check with Iman. I would not implement anything custom. It is fine to only keep the log for now, when the feature will be available we will add that. Maybe - just to remember this - place a comment referencing the DA174 spec as the way forward

tests/integration/application-charm/src/charm.py
deusebio: I discussed this with the other DP managers, and everyone was more inclined to handle everything in a single relation, without the need to decouple the lifecycle of users and resources. So feel free to resolve the comment with no further changes!

tests/integration/test_provider.py
deusebio: Yes, understood! Thank you!

deusebio:
> I believe it would be good to decouple the keyspace lifecycle from the user lifecycle in the test cham (unless we have strong reasons why they should be attached). This way we could also test that we only remove the user, and we make sure they are roles are still there (to make sure that the bug above is effectively fixed)

I don't think this applies anymore. Feel free to disregard this.

src/events/provider.py
BON4: I will ask Iman. However adding custom values to request/response is fairly easy. Here is the example from tests for data_interfaces v1:
```python
class ExtendedResponseModel(ResourceProviderModel):
    topsecret: ExtraSecretStr = Field(default=None)
    donttellanyone: ExtraSecretStr = Field(default=None)
    new_field: str | None = Field(default=None)
    new_field_req: str | None = Field(default=None)
    new_field2_req: str | None = Field(default=None)

class ExtendedRequirerCommonModel(RequirerCommonModel):
    new_field: str | None = Field(default=None)
```

imanenami: As @deusebio mentioned, this feature is available in DI v0.55 and later, and also will be ported to DI v1. You can see the open PR for v0 (which will be merged soon) [here](https://github.com/canonical/data-platform-libs/pull/242). Basically, you will have access to `raise_status`, `resolve_status` and `clear_statuses` methods on the provider to raise errors.

src/managers/database.py
Zvirovyi: I mean let's make:
```python
def all_valid_permissions() -> list[str]:
    return ["ALTER", ..., "CREATE"]
```
And then we'll be able to use it instead of this set directly just like:
```python
invalid = [p for p in perms if p.upper() not in [*self.all_valid_permissions(), "ALL"]]
```
This way we won't need `_valid_permissions` structure.

Not blocking though.

src/managers/database.py
```diff
@@ -112,6 +164,79 @@ def init_admin(self, password: str) -> None:
...
+        valid_ks = self.validate_identifier(ks)
+
+        cql = f"""
+        CREATE KEYSPACE IF NOT EXISTS {valid_ks}
+        WITH REPLICATION = {{'class': 'SimpleStrategy', 'replication_factor': %s}}
+        """
+
+        with self._session(
+            hosts=self.hosts,
+            auth_provider=self.auth_provider,
```
Zvirovyi, line 178: **suggestion** we can safely use `self._session() as session` here because `self.hosts` and `self.auth_provider` are used in `self._session` by default.
Same applies to other introduced functions.
Non-blocking.

Zvirovyi (approved):
Remarkable work! Awesome tests!

Image

deusebio (approved):
LGTM!

PR #23, opened 2025-10-15
[DPE-8322] Restart HA integration test & Continuous writes

Zvirovyi:
This PR:
* organizes all the integration helpers inside `helpers` folder;
* introduces continuous writes support based on multiprocessing lib;
* improves restart integration test to include HA tests requirements.


tests/integration/helpers/continuous_writes.py
```diff
@@ -0,0 +1,163 @@
...
+        position = self._get_max_position(hosts)
+
+        self.write_event.clear()
+        self.write_event.wait(self.force_timeout)
+
+        next_position = self._get_max_position(hosts)
+
+        assert next_position > position
+
+    def _init_keyspace(self, replication_factor: int) -> None:
```
deusebio, line 89: **question** shouldn't we check if the keyspace already exists? Can we make this idempotent

tests/integration/helpers/continuous_writes.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def _clear_keyspace(self) -> None:
+        self._cql_exec(f"DROP KEYSPACE IF EXISTS {self.keyspace_name}")
+
+    def _get_max_position(self, hosts: list[str] | None = None) -> int:
+        res = self._cql_exec("SELECT MAX(position) FROM test_table", hosts=hosts)
+        assert isinstance(res, ResultSet)
+        return res.all()[0][0]
+
+    def _assert_writes(self, hosts: list[str] | None = None) -> None:
```
deusebio, line 113: **niptick** it would seem more scalable to retrieve `min`, `max` and `count`. Then we can assert: `count == max -min`

tests/integration/helpers/continuous_writes.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class ContinuousWrites:
+    def __init__(self, keyspace_name: str = "test_keyspace", timeout: int = 600) -> None:
+        self.keyspace_name = keyspace_name
+        self.timeout = timeout
+        self.force_timeout = timeout * 2
+        self.stop_event = Event()
+        self.write_event = Event()
+        self.juju: jubilant.Juju | None = None
```
deusebio, line 24: **niptick,out-of-scope** I was a bit surprised to see `juju` here, since this is effectively a client application/script, that does not need to manage the deployment. I went into the code and I understood that effectively we use `juju` (beside some assertion) in the `connect_cql` function, to auto-infer the password and the hosts. 

I believe it would make more sense to: get the password and hosts outside of this class, and then feed these information into the class and used them in `connect` function. I believe it would be an "logical" improvement with respect to what we have now, but it is not critical, and it would probably require a bit of refactoring. I would leave it out of scope, but still create a ticket in the backlog if we will have some spare time

tests/integration/test_restart.py
```diff
@@ -0,0 +1,152 @@
...
+def test_freeze_process(
+    juju: jubilant.Juju, app_name: str, continuous_writes: ContinuousWrites
+) -> None:
+    continuous_writes.start(juju, app_name, [get_address(juju, app_name, 1)], replication_factor=3)
+
+    continuous_writes.assert_new_writes([get_address(juju, app_name, 0)])
+
+    send_control_signal(juju, f"{app_name}/0", "SIGSTOP")
+
+    sleep(10)
```
deusebio, line 81: **question** why is the sleep here lower than in other cases?

tests/integration/test_restart.py
```diff
@@ -0,0 +1,152 @@
...
+        delay=20,
+        timeout=1800,
+    )
+
+    continuous_writes.assert_new_writes([get_address(juju, app_name, 0)])
+    continuous_writes.assert_new_writes([get_address(juju, app_name, 1)])
+    continuous_writes.stop_and_assert_writes([get_address(juju, app_name, 2)])
+
+
+def send_control_signal(
```
deusebio, line 137: **niptick** should we place this in the `helpers`?

deusebio (approved):
**praise** I have very much appreciated the refactoring to place the business logic into helpers modules, and leave the tests rather concise with only the testing logic! well done!

Besides, everything looks good here! Nothing major, only some minor points/suggestions. Hence already approving. 

There are just a few tests failing in CI, not sure if it is flakiness or regressions, but please just make sure we have a full green CI before merging.

tests/integration/test_restart.py
Zvirovyi: This is used only in single test, so I don't think it viable to move it to the helpers.

tests/integration/test_restart.py
Zvirovyi: While in other cases we wait 60s to ensure Cassandra charm detection of service malfunction during update-status hook (that's being executed every 60s), here we just wait for other units to discover state of that unit.

tests/integration/helpers/continuous_writes.py
Zvirovyi: Keyspace is created when user executes `continuous_writes.start()` operation, and this operation shouldn't be executed multiple times. It's asserted in the top of `start` function.

tests/integration/helpers/continuous_writes.py
Zvirovyi: https://warthogs.atlassian.net/browse/DPE-8598?atlOrigin=eyJpIjoiMmJkYjUwZjk3NTI0NDY0N2JmMmExYjM1YmY4ZDFkZmYiLCJwIjoiaiJ9

Zvirovyi:
> There are just a few tests failing in CI, not sure if it is flakiness or regressions, but please just make sure we have a full green CI before merging.

Yes, flakiness of the restart test is indeed introduced by the [system_auth issue I've discovered yesterday](https://chat.canonical.com/canonical/pl/goddjqwuctns7cbgr957joed9y). I've started working on it.

tests/integration/helpers/cassandra.py
```diff
@@ -0,0 +1,219 @@
...
+def connect_cql(
+    juju: jubilant.Juju,
+    app_name: str,
+    hosts: list[str] | None = None,
+    username: str | None = None,
+    password: str | None = None,
+    keyspace: str | None = None,
+    client_ca: str | None = None,
+    timeout: float | None = None,
+) -> Generator[Session, None, None]:
```
theoctober19th, line 32: nit: add docstrings

theoctober19th (approved):
Superb work, @Zvirovyi! The refactoring is very sensible and I think the work here can even be taken as reference in refactoring integration test helpers in our other charms.

One thing I saw missing in some functions are the function doctsrings -- please add them wherever possible. On overall, looks good to me. 👍

PR #26, opened 2025-10-20
[DPE-8599] Repair management & high-availability for system_auth

Zvirovyi:
# Issue
Currently, Cassandra cluster deploys with a `system_auth` keyspace replication factor set to 1. It means that only single copy of all of the authentication system data is stored across the cluster. And it leads to the availability issues because authentication data can be easily corrupted (or inconsistencies occur) during scaling down & HA scenarios. Those issues already affect our integration testing.

# Overview
This PR:
- extends state with `AuthRepairState`, `ClusterContext.auth_repair`, `UnitContext.auth_repaired`;
- adds `update_auth_rf` custom event logic in CassandraEvents to handle `system_auth` keyspace replication factor scaling & repair.

# Notes
Scaling of system_auth rf was moved into distinct event due to specific conditions in which replication factor of keyspace can be changed: unit should be operational and no token movements (bootstrapping / decommissioning of the unit) should be ongoing.
While replication factor scaling on it's own can provide sufficient redundancy to partially avoid availability issues, full repair of this keyspace is required by-design to completely avoid `bad credentials` & `bad metadata detected` errors due to this keyspace inconsistency errors.

### DIFF
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -34,7 +34,13 @@ from tenacity import (
 from common.exceptions import BadSecretError, ExecError
 from core.config import CharmConfig
 from core.literals import CASSANDRA_ADMIN_USERNAME
-from core.state import DATA_STORAGE, PEER_RELATION, ApplicationState, UnitWorkloadState
+from core.state import (
+    DATA_STORAGE,
+    PEER_RELATION,
+    ApplicationState,
+    AuthRepairState,
+    UnitWorkloadState,
+)
 from core.statuses import Status
 from core.workload import WorkloadBase
 from managers.config import ConfigManager
@@ -295,25 +301,9 @@ class CassandraEvents(Object):
             return
         if self.config_manager.render_cassandra_config():
             self.restart()
-
-        if (
-            self.charm.unit.is_leader()
-            and self.state.cluster.auth_repairing
-            and all(unit.auth_repaired for unit in self.state.units)
-        ):
-            self.state.cluster.auth_repairing = False
-            self.state.unit.auth_repaired = False
-
-        if not self.charm.unit.is_leader():
-            if not self.state.unit.auth_repaired and self.state.cluster.auth_repairing:
-                if not self.state.unit.is_operational:
-                    logger.debug(
-                        "Deferring on_peer_relation_changed due to unit not being operational"
-                    )
-                    event.defer()
-                    return
-                self.node_manager.repair_auth()
-            self.state.unit.auth_repaired = self.state.cluster.auth_repairing
+        if not self._repair_auth(event):
+            event.defer()
+            return
 
     def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
         if not self.state.unit.is_ready:
@@ -333,6 +323,10 @@ class CassandraEvents(Object):
             self._recover_seeds()
             self.charm.on.update_auth_rf.emit()
 
+        if not self._repair_auth(event):
+            event.defer()
+            return
+
     def _on_leader_elected(self, event: LeaderElectedEvent) -> None:
         if not self.state.unit.is_ready:
             logger.debug("Exiting on_leader_elected due to unit not being ready")
@@ -343,6 +337,9 @@ class CassandraEvents(Object):
             return
         self._recover_seeds()
         self.charm.on.update_auth_rf.emit()
+        if not self._repair_auth(event):
+            event.defer()
+            return
 
     def _on_update_status(self, _: UpdateStatusEvent) -> None:
         if self.state.unit.is_operational and not self.workload.is_alive():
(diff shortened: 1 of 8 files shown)


deusebio:
The changes are overall fine, but we should just understand and fix the tests failures before merging.

src/managers/database.py
```diff
@@ -253,6 +253,14 @@ def validate_identifier(name: str) -> str:
             raise ValueError(f"Invalid CQL identifier: {name}")
         return name
 
+    def update_system_auth_replication_factor(self, replication_factor: int) -> None:
+        """Update replication factor of system_auth keyspace."""
+        with self._session() as session:
+            session.execute(
+                "ALTER KEYSPACE system_auth WITH replication = "
+                f"{{'class': 'SimpleStrategy', 'replication_factor': {replication_factor}}}"
+            )
```
marcoppenheimer, line 262: **question:** What might be the negative impacts of just calling this function and the `repair` on every update status?

Zvirovyi: Repair only fixes inconsistencies after altering a replication factor. Further ordinary read/writes will be automatically & correctly propagated to the quorum without need of repair.
So, even while repair of such small keyspace is fast, it still will be a waste of resources (operation takes about 10s).

Zvirovyi: Also: while that repair is inevitably required by the auth scaling logic, if we return to our previous dialogs, further "automatic repair" mechanisms like in update-status are out-of-scope and should be implemented distinctly.

marcoppenheimer: Ok, makes sense, I was wrong about repair. But what about just always checking if we need to up/down the replication factor on every event? That way we wouldn't need a lot of the state management in the PR, am I correct?

Something like:
```python
def _on_update_status():
    ...
    if replication_needs_changing():
        update_system_auth_replication_factor()
        repair_auth()
```

marcoppenheimer: I may be being VERY ignorant here so sorry if my suggestion is not possible!

Zvirovyi: Repair should be executed on all units, so state management will be needed anyways.

src/events/cassandra.py
```diff
@@ -435,3 +479,41 @@ def _on_storage_detaching(self, _: StorageDetachingEvent) -> None:
...
+            return
+        self.state.cluster.auth_repair = AuthRepairState.PENDING
+        if not self.state.unit.is_operational:
+            logger.debug("Deferring on_peer_relation_departed due to unit not being operational")
+            event.defer()
+            return
+        if (n := self.node_manager.active_cluster_nodes_count) != self.charm.app.planned_units():
+            event.defer()
+            return
+        self.database_manager.update_system_auth_replication_factor(min(n, 3))
```
deusebio, line 494: **out-of-scope** I was checking [here](https://stackoverflow.com/a/25892077) and they suggest to have the replication factor to the number of nodes, but this may also be overkill. Maybe it could be interesting to configure this via a config option with either `max_number` that can be a number or "all" (mean all the nodes). 

But this should be a different ticket, where we also agree on whether or not we want to have this feature and customization. 

@marcoppenheimer maybe worth just having a ticket in the backlog

deusebio (approved):
Looks good. The only thing is that most of the modifications in this PR actually deals with "Repair" management, rather than *just* scaling up the replication factor of system_auth. I would **just** provide a more explainable title for the PR/Jira something like:

`[DPE-8599] Repair management and system_auth high-availability`

src/events/cassandra.py
Zvirovyi: Replication factor of 3-5 is production-recommended value, higher value will be an overkill. [Source in official documentation](https://cassandra.apache.org/doc/stable/cassandra/managing/operating/security.html#password-authentication)

src/managers/config.py
```diff
@@ -279,6 +279,6 @@ def _cassandra_default_config() -> dict[str, Any]:
                 "cached_rows_warn_threshold": 2000,
             },
             "role_manager": "CassandraRoleManager",
-            "storage_compatibility_mode": "CASSANDRA_4",
```
welpaolo, line 282: Why this have been changed?

src/events/cassandra.py
```diff
@@ -435,3 +479,41 @@ def _on_storage_detaching(self, _: StorageDetachingEvent) -> None:
...
+
+    def _on_update_auth_rf(self, event: EventBase) -> None:
+        if not self.charm.unit.is_leader():
+            return
+        self.state.cluster.auth_repair = AuthRepairState.PENDING
+        if not self.state.unit.is_operational:
+            logger.debug("Deferring on_peer_relation_departed due to unit not being operational")
+            event.defer()
+            return
+        if (n := self.node_manager.active_cluster_nodes_count) != self.charm.app.planned_units():
```
welpaolo, line 491: Is this check in place to check that the cassandra app is not scaling up or down ?

welpaolo (approved):
LGTM. 
I left a couple of questions.

src/managers/config.py
Zvirovyi: This was made during my attempts to fix credentials outdated & bad metadata errors (that were caused by not doing repair). But I eventually left this change as we previously agreed with @marcoppenheimer that we should not support this feature. Source: [Gemini meeting transcription](https://docs.google.com/document/d/1QjWUX1hNxvTaPFr4e8IlyeESESeSu8K0dBCcn_MsvUc/edit?usp=sharing) (00:31:19)

src/events/cassandra.py
Zvirovyi: Yes, repairs cannot be done during any scaling.

PR #28, opened 2025-10-22
[MISC] General improvement for how-to docs

deusebio:
I have mostly improved the TLS guide to provide a bit more "charming" experience using actions, and using the `CA` certificate which is more conceptually correct, instead of the single client certificates. I have tested this and it works as expected. I also provided a bit more instruction to use the `charmed-cassandra` snap that we provide as part of the solution, also to connect as clients. 

I have also applied the usual convention of referring to "Apache <product>" and fixed a couple of typos coming from the Kafka documentation

### DIFF
--- a/docs/how-to/encryption.md
+++ b/docs/how-to/encryption.md
@@ -28,13 +28,13 @@ For an overview of available TLS certificate Provider charms and guidance on cho
 
 ## Relate the charms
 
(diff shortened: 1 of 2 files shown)


Zvirovyi (approved):
LGTM

BON4 (approved):
Looks good.

PR #30, opened 2025-10-24
[DPE-8634] Fix сassandra getting stuck when scaling up with TLS

BON4:
Fix an issue where Cassandra units would get stuck in the installing state when scaling up with TLS enabled.

Previously, when a new unit was added after TLS had been enabled, it could not successfully start because:

The new unit could not connect to the seed nodes (TLS handshake failed).
The charm deferred the certificate-available event indefinitely since the workload state was still `INSTALLING`.
As a result, the unit never updated its certificates nor completed the startup phase.

This change ensures that if the seeds are not yet reachable, the unit transitions to a `WAITING_FOR_START` state instead of remaining in `INSTALLING`.
This allows the charm to correctly defer and later handle the certificate-available event once the cluster becomes reachable.

Also, added new integration test, to specifically check this scenario.

### DIFF
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -193,6 +193,7 @@ class CassandraEvents(Object):
             return
 
         if not self._are_seeds_reachable:
+            self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START            
             logger.debug("Deferring subordinate on_start due to seeds not being ready")
             event.defer()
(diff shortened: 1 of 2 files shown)


Zvirovyi (approved):
Nice catch!

deusebio (approved):
Nice!!!! Great job at solving this so swiftly!

PR #31, opened 2025-10-27
[DPE-8731] Add unit kill ha test

BON4:
Add test: kill primary, ensure cluster works.
Should be merged after #36

### DIFF
--- a/tests/integration/test_kill.py
+++ b/tests/integration/test_kill.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+from pathlib import Path
+
+import jubilant
+
+from integration.helpers.cassandra import OPERATOR_PASSWORD
+from integration.helpers.continuous_writes import ContinuousWrites
+from integration.helpers.juju import (
+    app_secret_extract,
+    get_hosts,
+    get_leader_unit,
+    get_non_leader_units,
+    scale_sequentially_to,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def test_deploy(juju: jubilant.Juju, cassandra_charm: Path, app_name: str) -> None:
+    juju.deploy(
+        cassandra_charm,
+        app=app_name,
+        config={"profile": "testing"},
+        num_units=3,
+    )
+    juju.wait(
+        ready=lambda status: jubilant.all_agents_idle(status) and jubilant.all_active(status),
+        delay=20,
+        timeout=2400,
+    )
+
+
+def test_kill_primary(
+    juju: jubilant.Juju, app_name: str, continuous_writes: ContinuousWrites
+) -> None:
+    continuous_writes.start(
+        hosts=get_hosts(juju, app_name),
+        password=app_secret_extract(juju, app_name, OPERATOR_PASSWORD),
+        replication_factor=3,
+    )
+
+    leader, _ = get_leader_unit(juju, app_name)
+    kill_unit(juju, leader)
+
+    juju.wait(
+        ready=lambda status: jubilant.all_agents_idle(status) and jubilant.all_active(status),
+        delay=20,
+        timeout=1800,
+    )
+    continuous_writes.assert_new_writes()
+
+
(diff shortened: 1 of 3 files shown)


Zvirovyi (changes requested):
1. Let's assert cluster health after switchover by doing write-read test.
2. Can we also add test for removing leader unit altogether?

tests/integration/test_charm.py
```diff
@@ -74,3 +80,29 @@ def test_write_primary_read_secondary(juju: jubilant.Juju, app_name: str) -> Non
...
+        ),
+        delay=5,
+        timeout=1800,
+    )
+
+    sleep(REELECTION_TIME)
+
+    for attempt in Retrying(wait=wait_fixed(30), stop=stop_after_delay(300), reraise=True):
+        with attempt:
+            new_leader, new_leader_status = get_leader_unit(juju, app_name)
```
deusebio: I don't think that the leader means the Juju leader, but rather for the workload (unless the workload leader corresponds with the juju leader). Is there any concept of keyspace leader as much as it is done for Kafka, e.g. [here](https://github.com/canonical/kafka-operator/blob/main/tests/integration/helpers/ha.py#L47)?

The idea of this test is that we kill the most important unit, trigger a re-election of the leader/master (for the workload), and check after some time that we have a new leader

deusebio (changes requested):
I'm not sure the current test is what is in scope for the HA tests. We should not be testing Juju, but the workload re-election (if there can be such a thing)

tests/integration/test_charm.py
Zvirovyi: Juju leader is the most important unit after cluster deployment as it was first to hold all the keyspaces. And only later data will be replicated to the other units as they join one-by-one.

deusebio: What happens when the node going down is the seed node? Do we have a re-election process?

Zvirovyi: Yes, after [seed node management PR](https://github.com/canonical/cassandra-operator/pull/13).

deusebio: All right! Then I believe it makes sense to test that the seed node is then switched after killing the leader unit. Can we add this? This can be interpreted as the re-election process. Beside verifying that we have a new juju leader, we should also verify that this change is processed by the charm and the seed node is moved (since the old one died)

@BON4

deusebio:
> 1. Let's assert cluster health after switchover by doing write-read test.

Yes, I agree with @Zvirovyi . I believe this makes sense!

> 2. Can we also add test for removing leader unit altogether?

I believe we already have the tests when removing the leader, see [here](https://github.com/canonical/cassandra-operator/blob/main/tests/integration/test_scaling.py#L97). Is this what you meant?

Zvirovyi:
> I believe we already have the tests when removing the leader, see [here](https://github.com/canonical/cassandra-operator/blob/main/tests/integration/test_scaling.py#L97). Is this what you meant?

Then do we really need this PR? @deusebio

Zvirovyi:
UPD: this PR was reworked, all the comments above are not relevant.

marcoppenheimer (approved):

tests/integration/test_kill.py
```diff
@@ -0,0 +1,74 @@
...
+
+    leader, _ = get_leader_unit(juju, app_name)
+    kill_unit(juju, leader)
+
+    juju.wait(
+        ready=lambda status: jubilant.all_agents_idle(status) and jubilant.all_active(status),
+        delay=20,
+        timeout=1800,
+    )
+    continuous_writes.assert_new_writes()
```
deusebio, line 61: **todo** can we check that we have updated the seed node? I'm particular keen to make sure that [this use case](https://chat.canonical.com/canonical/pl/3abayug7t3gsjeby1bbbqiog7y) is handled properly

deusebio (commented):
There is just one particular use-case I believe we should cover

zmraul (approved):
thanks! test setup seems fine, but I'm unsure of the exact checks needed for Cassandra. I'll trust the other reviewers with that.

marcoppenheimer (approved):

PR #34, opened 2025-10-30
[DPE-8612] Ha tests network

BON4:
Following tests are added:
1. network cut (no IP Change)
2. network change (network cut IP change)


tests/integration/test_full_cluster_restart.py
Zvirovyi, line 1: **todo** we've already renamed this test to `test_restart.py`

tests/integration/helpers/ha.py
```diff
@@ -0,0 +1,86 @@
...
+    except subprocess.CalledProcessError:
+        # Ignore if the interface was already overridden.
+        pass
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.egress=0kbit"
+    subprocess.check_call(limit_set_command.split())
+    limit_set_command = f"lxc config device set {machine_name} eth0 limits.ingress=1kbit"
+    subprocess.check_call(limit_set_command.split())
+
+
+def network_release(machine_name: str) -> None:
```
Zvirovyi, line 28: **todo** Let's make it to the single function like
```python
def network_throttle(machine_name: str, limits: int | None = None) -> None:
```
If `limits is None`, then we reset limits.

Zvirovyi (changes requested):
Nice work! Just a couple of comments to resolve

tests/integration/helpers/ha.py
deusebio: Uhm, the current implementation is consistent with what we have in Kafka. Although I believe it would have been a sane comment, I believe this is a minor point and it probably better to keep consistency

tests/integration/test_network.py
```diff
@@ -0,0 +1,182 @@
...
+        ready=lambda status: jubilant.all_agents_idle(status) and jubilant.all_active(status),
+        delay=20,
+        timeout=1800,
+    )
+
+
+def test_network_cut_without_ip_change(
+    juju: jubilant.Juju, continuous_writes: ContinuousWrites, app_name: str
+) -> None:
+    stoped_unit_name, stoped_unit_status = get_leader_unit(juju, app_name)
```
deusebio: **niptick** 
```suggestion
    stopped_unit_name, stopped_unit_status = get_leader_unit(juju, app_name)
```

Did you mean "stopped"? Also, maybe I would called it "detached" or "cut", since the network cut is effectively cutting the app/unit/machine out of the cluster

deusebio (approved):
Looks good! Just a niptick to rename the variable (while fixing the spelling)

Thanks

tests/integration/test_restart.py
Zvirovyi, line 1: Is this an old version of test?

Zvirovyi (changes requested):

tests/integration/test_restart.py
BON4: Left after merging. Fixed.

Zvirovyi (approved):
LGTM!

PR #36, opened 2025-11-12
[DPE-8883] Handle forceful unit removals

Zvirovyi:
This PR introduces handling of forceful unit removal, after which already removed Juju units still persist in the Cassandra cluster. Check for dead unknown Cassandra nodes is added in the update status event hook utilizing `nodetool status` & `nodetool removenode` commands.

### DIFF
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -352,16 +352,22 @@ class CassandraEvents(Object):
             self.restart()
             return
 
-        if self.state.unit.is_operational and self._update_network_address():
+        if not self.state.unit.is_operational:
+            logger.debug("Exiting on_update_status due to unit not being operational")
+            return
+
+        if self._update_network_address():
             if self.charm.unit.is_leader():
                 self.state.seed_units = self.state.unit
-
             self.config_manager.render_cassandra_config(
                 listen_address=self.state.unit.ip,
                 seeds=self.state.cluster.seeds,
             )
-
             self.restart()
+            return
+
+        if self.charm.unit.is_leader():
+            self.node_manager.remove_bad_nodes([unit.ip for unit in self.state.units])
 
     def _on_collect_unit_status(self, event: CollectStatusEvent) -> None:
         try:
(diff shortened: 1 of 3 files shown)


src/events/cassandra.py
```diff
@@ -352,16 +352,22 @@ def _on_update_status(self, _: UpdateStatusEvent) -> None:
...
-
             self.config_manager.render_cassandra_config(
                 listen_address=self.state.unit.ip,
                 seeds=self.state.cluster.seeds,
             )
-
             self.restart()
+            return
+
+        if self.charm.unit.is_leader():
```
welpaolo, line 369: Is there any case where the removal of nodes can cause data loss ? 
Is there any scenario where this can happen and the unit has not be removed (scale down) ?

welpaolo (commented):
LGTM. I left a couple of questions to understand some dynamics

src/events/cassandra.py
Zvirovyi: 1. Unit will not be removed from the Cassandra cluster if it was removed forcefully and / or `storage-departed` hook didn't fire on departing unit.
2. It can cause data loss if there are keyspaces with replication factor 1 held by departing unit. That shouldn't happen typically as production recommendations are to keep it at 3.

welpaolo (approved):
LGTM

src/managers/node.py
```diff
@@ -101,14 +102,38 @@ def decommission(self) -> None:
...
+
+        for down_node in down_nodes:
+            ip, host_id = down_node
+            if ip in good_node_ips:
+                continue
+            try:
+                self._workload.exec([_NODETOOL, "removenode", host_id])
+                logger.info(f"Removed bad node {ip} from Cassandra cluster")
+            except ExecError:
+                logger.error(f"Failed to remove bad node {ip} from Cassandra cluster")
```
imanenami, line 138: suggestion: can we add a unit test for this method, especially for validating the parsing logic?

Zvirovyi: It was a great idea, I've even fixed a bug while writing this test! While original regex was okay for handling ordinary cases such as:
`DN  10.222.10.185  190.05 KiB  16      52.0%             9e969d6a-10ce-4cfb-b537-a333f5990158  rack1`
there may be situations when nodes can appear with no `Load` column like:
`DN  10.222.10.185  ?  16      52.0%             9e969d6a-10ce-4cfb-b537-a333f5990158  rack1`
and it would've raise an exception. Thanks, and done.

imanenami (approved):

zmraul (approved):
Thanks!

PR #37, opened 2025-11-13
Improve bootstrapping logic

Zvirovyi:
Currently, Cassandra charm tries to run `nodetool bootstrap resume` command when it encounters `Bootstrap state: NEEDS_BOOTSTRAP` state during unit bootstrap. Such process will fail with "bootstrap resume command disabled" error, although right decision in this case will be to just wait for it to automatically start Cassandra bootstrap process. This PR fixes this to make logic right and to avoid blocked status flakiness during cluster deployments.


imanenami (approved):
LGTM, thanks.

marcoppenheimer (approved):

PR #39, opened 2025-11-17
[DPE-8884] Add charm refresh

BON4:
This PR introduces seamless minor version upgrades for **charmed Cassandra**, with zero downtime.

**Key changes included:**

* Added `MachinesCassandraRefresh` event handler to manage Cassandra-specific upgrade logic
* Introduced `UpgradesManager` to integrate `charm-refresh v3` functionality with advanced status handling
* Modified `CassandraWorkload` to support:

  * Installing a specific snap revision rather than only the latest
  * Reading snap revisions from the `refresh_versions` file instead of hard-coded values
* Updated critical event handlers to pause or defer operations while upgrades are running, following the spec guidelines
* Added unit tests covering Cassandra-specific upgrade logic (excluding charm-refresh core functionality)
* Modified Spread configuration to exclude `juju 3.5` tests, since it is no longer avaliable at snapstore.

**Not included in this PR:**

* Full integration testing
* Edge-case scenarios

Integration tests will be added in separate PRs, once charm versions with `charm-refresh v3` support are available, including tests with newer snap revisions.


src/core/workload.py
```diff
@@ -99,6 +99,13 @@ def install(self) -> None:
         """Install Cassandra."""
         pass
 
+    @property
+    @abstractmethod
+    def installed(self) -> bool:
+        """Whether Cassandra is installed."""
+        pass
```
imanenami, line 106: question: why `pass`? you can check for snap installed :thinking:

src/events/cassandra.py
```diff
@@ -402,8 +432,37 @@ def _on_collect_unit_status(self, event: CollectStatusEvent) -> None:
...
+            return False
+
+        refresh_status = self.refresh_manager.unit_status_lower_priority(
+            workload_is_running=self.workload.is_alive,
+        )
+        if refresh_status:
+            event.add_status(refresh_status)
+            return True
+
+        return False
```
marcoppenheimer: **nitpick:** Could we combine both of these in to a single method, and just pass `priority: Literal["high", "low"]` (or an enum) as an arg?

src/events/refresh.py
```diff
@@ -46,17 +46,20 @@ def run_pre_refresh_checks_after_1_unit_refreshed(self) -> None:  # type: ignore
...
-
-        for host in [u.ip for u in self._state.units]:
-            if not self._node_manager.is_healthy(host, retry=True):
-                raise charm_refresh.PrecheckFailed("Cluster is not healthy")
+        if not self._workload.installed:
+            raise charm_refresh.PrecheckFailed("Current unit workload is not installed")
+
+        for unit in self._state.units:
+            if not unit.is_ready:
+                continue
```
BON4, line 54: What the reason for this change? Why should we not check if unit is not ready?

src/managers/node.py
```diff
@@ -32,15 +34,31 @@ class NodeManager:
...
+            return _check()
+
+        for attempt in Retrying(
+            wait=wait_exponential(),
+            stop=stop_after_delay(timeout),
+            reraise=False,
+        ):
+            with attempt:
+                if _check():
+                    return True
```
marcoppenheimer: **nitpick:** I don't think we need the nested function here, we can just set the number of retries to `1` is `retry=False`

src/events/refresh.py
Zvirovyi: During scaling up, multiple units can stay in `not is_ready` state as they will be paused based on `self.refresh_manager.workload_allowed_to_start` value. So we shouldn't assert their health as they're not part of the cluster yet. And when refresh is finished, condition above will succeed and all of the remaining units will be allowed to join in.

Zvirovyi (approved):
Amazing work!

src/charm.py
```diff
@@ -115,8 +136,17 @@ def __init__(self, *args):
             log_slots=[f"{SNAP_NAME}:logs"],
         )
 
+        if (
+            self.refresh_manager.is_initialized
+            and not self.refresh_manager.next_unit_allowed_to_refresh
```
imanenami, line 141: suggestion: since these twin checks are usually done together, it's better to abstract them away in some form of state/readiness check either in the context or the refresh manager.

src/charm.py
```diff
@@ -115,8 +136,17 @@ def __init__(self, *args):
...
+            and self.workload.installed
+        ):
+            # Only proceed if snap is installed (avoids KeyError during initial deployment)
+            self.refresh.post_snap_refresh(self.refresh_manager)
+
     def _on_bootstrap(self, event: EventBase) -> None:
         if self._handle_starting_state(event):
+            logger.info("Deferring _on_bootstrap")
```
imanenami: suggestion: debug log might be better, same for other added logs

zmraul: **todo**: These logs should be debug or removed

src/charm.py
```diff
@@ -133,10 +163,12 @@ def _on_bootstrap(self, event: EventBase) -> None:
             event.defer()
             return
 
+        logger.info("Trying to check heath in _on_bootstrap")
```
zmraul: todo: These logs should be debug or removed

src/charm.py
```diff
@@ -133,10 +163,12 @@ def _on_bootstrap(self, event: EventBase) -> None:
...
         if not self.node_manager.is_healthy(ip=self.state.unit.ip):
             logger.debug("Deferring on_bootstrap due to workload not being healthy yet")
             event.defer()
             return
+        logger.info("Heath in _on_bootstrap is good")
```
zmraul: todo: These logs should be debug or removed

src/charm.py
```diff
@@ -53,6 +64,15 @@ def __init__(self, *args):
...
         self.tls_manager = TLSManager(workload=self.workload)
 
+        self.refresh = MachinesRefresh(
+            workload_name="charmed-cassandra",
+            charm_name="cassandra",
+            _state=self.state,
+            _workload=self.workload,
+            _node_manager=self.node_manager,
+        )
+        self.refresh_manager = RefreshManager(self.refresh)
```
zmraul, line 74: **suggestion**: I like the separation of concerns here in theory. In practice I'm not sure it works. Every call to `refresh` methods should be generally be preceded by a `refresh_manager` check for `is_initialized`. I think this defeats the purpose of splitting them and creates a weird pattern of interaction. Refresh doesn't play well into the charm layout we use so I would rather keep the logic in the event handler altogether.

Not a blocker, and happy to be disagreed with here.

src/events/cassandra.py
```diff
@@ -185,13 +198,15 @@ def _start_leader_setup_auth(self) -> None:
             truststore_password=self.state.unit.truststore_password,
         )
         self.workload.start()
+        logger.info("Trying to check heath in _start_leader_setup_auth")
```
imanenami: I think the logs here should also be debug.

src/events/cassandra.py
```diff
@@ -402,8 +432,37 @@ def _on_collect_unit_status(self, event: CollectStatusEvent) -> None:
...
+    def _handle_refresh_high_priority(self, event: CollectStatusEvent) -> bool:
+        if not self.refresh_manager.is_initialized:
+            return False
+
+        if self.refresh_manager.unit_status_higher_priority:
+            event.add_status(self.refresh_manager.unit_status_higher_priority)
+            return True
+
+        if status := self.refresh_manager.unit_status_lower_priority():
+            event.add_status(status)
```
zmraul: **todo**: I don't think this should be here. It is covered by `_handle_refresh_manager_low_priority`

src/managers/node.py
```diff
@@ -32,15 +34,31 @@ class NodeManager:
...
+            return True
+
+        if not retry:
+            return _check()
+
+        for attempt in Retrying(
+            wait=wait_exponential(),
+            stop=stop_after_delay(timeout),
+            reraise=False,
+        ):
```
imanenami, line 45: suggestion: the `is_healthy` method could be decorated with `tenacity` decorators for leaner syntax...

imanenami (approved):
Great work overall! I have no major concern, except some possible improvement on the level of some logs. In general, I think anything which is related to developers should be logged at DEBUG level.

Also, I recommend to add a TODO in the integration tests so that once there is a release, it's not forgotten to amend them :)

src/events/refresh.py
```diff
@@ -0,0 +1,121 @@
...
+        old_workload_version: str,
+        new_workload_version: str,
+    ) -> bool:
+        """Check charm and workload version compatibility."""
+        return super().is_compatible(
+            old_charm_version=old_charm_version,
+            new_charm_version=new_charm_version,
+            old_workload_version=old_workload_version,
+            new_workload_version=new_workload_version,
+        )
```
zmraul, line 43: **todo**: ~Optional, but good to keep in mind.~ The current snap/charm is only on track `5`, so there is no major version change happening. However, you could add a placeholder for future reference like we have [in Kafka](https://github.com/canonical/kafka-operator/blob/main/src/events/refresh.py#L109) to document upgrade paths, see the end of that kafka function.

edit: moving this to **todo**. I checked again, and this can be added. Checks to avoid downgrades are still useful, like [this one](https://github.com/canonical/kafka-operator/blob/main/src/events/refresh.py#L136-L141)

zmraul (changes requested):
Great thanks! Left some comments.

(on the logs, generally search for `logger.info` added by this PR, there are more on other files)

src/core/workload.py
Zvirovyi: `@abstractmethod`

src/charm.py
Zvirovyi: We got only 1 occurrence of this check, so I don't think it's needed right now.

src/charm.py
Zvirovyi: From [docs](https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/unit-tearing-down/):

> ```python
> class PostgreSQLCharm(ops.CharmBase):
>     def __init__(self, *args):
>         # [...]
> 
>         try:
>             self.refresh = charm_refresh.Machines(
>                 MachinesPostgreSQLRefresh(
>                     workload_name="PostgreSQL",
>                     charm_name="postgresql",
>                 )
>             )
>         except charm_refresh.PeerRelationNotReady:
>             # [...]
>         except charm_refresh.UnitTearingDown:
>             
>             self.unit.status = ops.MaintenanceStatus("Tearing down")
>             snap.uninstall()
>             sys.exit() 
> ```
> This may be incompatible with your charm code or charm code dependencies that depend on executing code during specific Juju events. In those cases, a more complex approach may be needed

We need to at least execute storage detached hook during scaling down.

@BON4, what do you think?

Zvirovyi: @zmraul 
Or you mean moving whole refresh to CassandraEvents?

src/events/cassandra.py
Zvirovyi: From [docs](https://canonical-charm-refresh.readthedocs-hosted.com/latest/add-to-charm/status/#implementation):
> [ops' collect status event](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.CollectStatusEvent) must not be used.

And we're using it 🤪. So, we should add it here for the purpose of not overriding it with our own status collected here.

src/events/refresh.py
Zvirovyi: `super().is_compatible` already handles major versions & downgrade incompatibilities:
```python
    @staticmethod
    def _is_charm_version_compatible(*, old: CharmVersion, new: CharmVersion):
        """Check that new charm version is higher than old and that major versions are identical

        TODO talk about intermediate charms

        TODO talk about recommendation to not support charm code downgrade
        """
        if not (old.released and new.released):
            # Unreleased charms contain changes that do not affect the version number
            # Those changes could affect compatability
            return False
        if old.track != new.track:
            return False
        if old.major != new.major:
            return False
        # By default, charm code downgrades are not supported (rollbacks are supported)
        return new >= old
```

And as per @marcoppenheimer we will not support major version upgrades. So we don't need to change anything here.

src/charm.py
zmraul: Yes sorry, I had the whole conversation on my head and didn't explain. I mean having refresh_manager be a wrapper for `charm_refresh.Machines` and handling the exceptions there.

I think only `events/refresh.py` is needed, and we don't need the two classes, but just have the `try-except` for `PeerRelationNotReady` and `UnitTearingDown` in the init directly But is more of a design thing than a strong requirement.

src/managers/node.py
Zvirovyi: Indeed this function can be improved. Let's maybe split it into the `is_healthy` & `is_healthy_now` functions or something similar.

src/events/cassandra.py
zmraul: Yes, but my understanding is that we can integrate into collect status:
   1) If there is high priority from refresh, set that one and return
   2) Handle the charm statuses. This can and should override the low priority from refresh
   3) Lastly, if there are low priority and the charm doesn't need to set a status, set the low priority refresh

[kafka implementation](https://github.com/canonical/kafka-operator/blob/main/src/charm.py#L213)

src/events/refresh.py
zmraul: good to know, thanks. I guess we can remove those initial checks from Kafka then :)

src/events/cassandra.py
Zvirovyi: We already do 1. I was talking about [vanilla Ops CollectStatusEvent](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.CollectStatusEvent), not the Kafka's custom one:
> If any statuses were added by the event handler using [add_status()](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.CollectStatusEvent.add_status), the framework will choose the highest-priority status and set that as the status.
> 
> If there are multiple statuses with the same priority, the first one added wins (and if an event is observed multiple times, the handlers are called in the order they were observed).

And we already add the refresh low priority status as latest one [here](https://github.com/canonical/cassandra-operator/blob/8710b1086a604f0aaa86fb3888bd383cf6eccae8/src/events/cassandra.py#L434). So it provides almost the same logic like in 2 & 3.

src/charm.py
Zvirovyi: Do you mean that it's better to not have a distinct class for handling 2 exceptions and handle them right away in `charm.py` or `events/cassandra.py`?

> I think only `events/refresh.py` is needed

What do you mean?

> and we don't need the two classes

Are you talking about `Refresh` & `MachinesRefresh`?

Zvirovyi: We keep event handlers fully separate from each other. The only [approach](https://github.com/canonical/cassandra-operator/pull/9#discussion_r2208682316) we agreed upon was to place shared code in general `charm.py` file and then passthrough it down to all the handlers, but never do cross-links between two events. `MachinesRefresh` = event handler, because it defines events, while `RefreshManager` just provides info about refresh state.

marcoppenheimer (approved):

zmraul (approved):

PR #42, opened 2025-12-15
[DPE-9057] Seed nodes management

Zvirovyi:
**Depends on #43!**

This PR:
* implements [DA231 - Cassandra Seed Nodes Management specification](https://docs.google.com/document/d/1kikME-GvbUUeGjlyl5rvfforzfw7rw4sxAuXCYurQlM/edit?usp=sharing);
* improves hash support for the RelationState to support sets of UnitContexts.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -105,6 +105,7 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
             setup_internal_certificates=self.setup_internal_certificates,
             read_auth_secret=self.read_auth_secret,
             restart=self.restart,
+            reconcile_seeds=self.reconcile_seeds,
         )
 
         self.tls_events = TLSEvents(
@@ -116,6 +117,7 @@ class CassandraCharm(TypedCharmBase[CharmConfig]):
             tls_manager=self.tls_manager,
             setup_internal_certificates=self.setup_internal_certificates,
             restart=self.restart,
+            reconcile_seeds=self.reconcile_seeds,
         )
 
         self.provider_events = ProviderEvents(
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -273,6 +275,36 @@ def setup_internal_certificates(self, sans: Sans) -> bool:
 
         return True
 
+    def reconcile_seeds(self, ensure_seed: bool = False) -> bool:
+        """Reconcile seeds accordingly to DA231. Scale up to 3 units.
```
imanenami, line 279: todo: We need to move this spec to the DA folder, and assign a new identifier to it. Also the spec requires formal approval.

src/charm.py
```diff
@@ -105,6 +105,7 @@ def __init__(self, *args):
             setup_internal_certificates=self.setup_internal_certificates,
             read_auth_secret=self.read_auth_secret,
             restart=self.restart,
+            reconcile_seeds=self.reconcile_seeds,
```
imanenami, line 108: suggestion: using the event-driven approach for reconciliation seems like an anti-pattern. I think we should refactor this PR. and run the reconciliation on collect status, which is run after all hooks...

imanenami:
@marcoppenheimer @zmraul 

Let's finalize the spec: https://docs.google.com/document/d/1kikME-GvbUUeGjlyl5rvfforzfw7rw4sxAuXCYurQlM/edit?usp=sharing

The implementation looks fine, the only major change required imo is the fact that we should develop an "actual" reconcile method, instead of cherry-picking events... I'll finalize this PR once you take a look at the spec and this initial implementation and provide feedback...

zmraul (commented):
Similar concerns as @imanenami. As it is right now, there is no seed removal at scale down or other situations like leader-changed are not handled. Moving to a reconcile pattern will be best here.

PR #44, opened 2026-01-19
chore: switch default profile to testing

zmraul:
Current default is "production" which will use half of all available memory. This is not right specially in the context of LXD, since memory constraints need a specific config to be set.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -4,7 +4,7 @@
 options:
   profile:
     type: string
-    default: production
+    default: testing
     description: |
       Profile represents the scope of deployment, and used to tune resource allocation.
       Allowed values are: `production` and `testing`.


imanenami (approved):

imanenami:
I suggest to update the rust version in charmcraft.yaml to fix the build issue, otherwise the release would fail :)

marcoppenheimer (approved):

marcoppenheimer (changes requested):
Actually, I don't think we should do this.
Let's keep the default being `production`, and ensure that during tests + local deployments, we set the `testing` profile.

As an FYI, Cassandra uses this formula for allocating JVM heap size:

```
max(min(1/2 ram, 1024MB), min(1/4 ram, 8GB)
```

imanenami (commented):
Revoking approval as I agree with Marc. We should use `production` profile as default as it's the case in Kafka, Opensearch  and other DP charms, and it makes sense. Think we can close this PR :)

marcoppenheimer:
While I agree in practice, the strategy is to have `production` as default across most of our products. Let's close this, thanks.

PR #45, opened 2026-02-10
[DPE-9337] docs: Bootstrap docs for Cassandra

izmalk:
Bootstrap the Cassandra RTD docs in the repository.

On top of the branch PR, I've set up an RTD project to publish the `rtd-docs` branch for now (might need SSO login): https://canonical-charmed-cassandra.readthedocs-hosted.com/en/rtd-docs/

The `Check for removed URLs / build-docs (pull_request)` test is supposed to be broken until the first push to the main branch.


marcoppenheimer (approved):
Looks good, thank you! I suggest we add the 'dont run tests if only docs' change we have with Kafka too.

izmalk:
> I suggest we add the 'dont run tests if only docs' change we have with Kafka too.

It's already implemented in this PR. But we need to add the GItHub workflows as well as the `.readthedocs.yaml` config, and they are both outside of the `only docs` scope (which is defined as the `docs` directory). That is why this PR runs non-docs tests.

imanenami (approved):
LGTM 🚀

PR #46, opened 2026-02-13
[DPE-9436] fix: build & CI issues

imanenami:
## Description

- Bumps rust version to latest stable in builds
- Due to `microk8s` strict mode incompatibility with newer kernel version, switches to non-strict mode for COS tests.
- Some enhancement in `spread` configuration to avoid unnecessary bootstrap of a k8s controller where not needed.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -97,10 +97,11 @@ suites:
 path: /root/spread_project
 
 kill-timeout: 1h
+warn-timeout: 1m
 environment:
   PATH: $PATH:$(pipx environment --value PIPX_BIN_DIR)
   CONCIERGE_JUJU_CHANNEL/juju36: 3.6/stable
-  MICROK8S_CHANNEL: "1.28-strict/stable"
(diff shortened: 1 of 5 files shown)


spread.yaml
```diff
@@ -120,6 +122,28 @@ prepare-each: |
   # `concierge prepare` needs to be run for each spread job in case Juju version changed
   concierge prepare --trace
 
+  if [[ -v USES_K8S ]]; then
```
zmraul, line 125: Can't all of this be specified in concierge instead, and use canonical k9s?

imanenami: Only `test_cos` requires k8s bootstrap, so no. Prefer to use microk8s here which is more tested with the COS stack.

Gu1nness (approved):
I would probably do it in two steps, one which is concierge with bootstrap set to False, and then in your spread, do the second part (i.e move the creds to the right place + bootstrap, but this works as well!

zmraul (approved):

PR #59, opened 2026-04-06
[DPE-9741] feat: enable JMX remote management interface

imanenami:
## Description

This PR enables password-protected nodetool commands and enables remote JMX access, which is a prerequisite for backups implementation.


src/core/literals.py
```diff
@@ -5,3 +5,4 @@
 """Common literals in the charm."""
 
 CASSANDRA_ADMIN_USERNAME = "operator"
+NODETOOL_USERNAME = "charmed-operator"
```
marcoppenheimer, line 8: **nitpick:** Is there harm in having both usernames the same, with the same secret password?

src/managers/config.py
```diff
@@ -99,9 +99,11 @@ def render_env(self, cassandra_limit_memory_mb: int | None) -> bool:
             whether config was changed.
         """
         content = self._render_env(
-            self._map_env(self.workload.cassandra_paths.env.read_text().split("\n"))
+            {"LOCAL_JMX": "no"}
```
marcoppenheimer, line 102: **fluff:** `no` 😆

src/managers/node.py
```diff
@@ -31,8 +31,29 @@ class GossipNode:
...
+    @property
+    def password(self) -> str:
+        """Return the current password being used with nodetool."""
+        return self._password
+
+    @password.setter
+    def password(self, value: str) -> None:
+        self._password = value
+
+    def nodetool_exec(
```
marcoppenheimer, line 48: **praise:** Nice, good addition.

tests/unit/test_node_manager.py
```diff
@@ -41,21 +41,47 @@ def test_remove_bad_nodes():
...
 
     workload.exec.return_value = (status_one_active, None)
     node_manager.remove_bad_nodes(good_node_ips=["10.222.10.92", "10.222.10.254"])
     assert workload.exec.call_count == 3
     workload.exec.assert_has_calls(
         calls=[
-            call([NODETOOL, "status"]),
-            call([NODETOOL, "removenode", "8f6406bd-96f5-43c3-9433-8439773b7495"]),
-            call([NODETOOL, "removenode", "1a820086-663b-496f-87b7-d6ccf99df077"]),
+            call(
```
marcoppenheimer, line 58: **thought:** Not for this PR as I think it's best to not change the existing work too much - but making note that I tend to dislike the `call` tests as they end up fairly coupled. Nicer to check `is 'password' in the call args list` or something IMO

marcoppenheimer (approved):
Good changes!

src/core/literals.py
imanenami: I think it's better to keep them separate, as they serve two different purposes on different interfaces...

tests/unit/test_node_manager.py
imanenami: You're right, this is a bad test. I'll fix this test since it's somehow related to this PR.

src/managers/node.py
```diff
@@ -31,8 +31,29 @@ class GossipNode:
...
+        self, *args: str, host: str | None = None, suppress_error_log: bool = False
+    ) -> str:
+        """Execute a nodetool command."""
+        cmd_args = [NODETOOL, "-u", self._username, "-pw", self._password]
+        if host:
+            cmd_args += ["-h", host]
+        stdout, _ = self._workload.exec([*cmd_args, *args], suppress_error_log=suppress_error_log)
+        return stdout
```
zmraul, line 56: **question**: Why not wrapping this around something like `workload.run_snap_command(<bin_keyword>)` like we do in Kafka? We are now copying `charmed-cassandra.nodetool` on several places. Not a blocker though.

zmraul (approved):
LGTM, thanks!

src/managers/node.py
imanenami: I think what `nodetool_exec` achieves is exactly what you mentioned. I agree that before this refactoring, it was not DRY as you mentioned.

PR #60, opened 2026-04-07
[DPE-9742] feat: add SSH manager

imanenami:
This PR adds a simple SSH manager to enable SSH intercom between nodes + a reconciliation mechanism.


src/events/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+import abc
+
+
+class Reconciliable(abc.ABC):
+    """Mixin ABC for reconciliable event handlers."""
+
+    def reconcile(self) -> None:
+        """Reconcile this event handler."""
+        pass
```
marcoppenheimer, line 15: **question:** What value does this provide over just adding a `reconcile` method?

src/events/cassandra.py
```diff
@@ -116,13 +122,20 @@ def _acquire_operator_password(self) -> str:
...
+    def _on_start(self, event: StartEvent) -> None:  # noqa: C901
         if not self.state.peer_relation:
             event.defer()
             return
 
         self._update_network_address()
 
+        if not self.ssh_manager.public_key:
+            self.ssh_manager.keygen()
+            event.defer()
```
marcoppenheimer, line 134: **question:** What does the defer() here achieve?

src/managers/ssh.py
```diff
@@ -0,0 +1,83 @@
...
+
+        This operation is idempotent.
+        """
+        current = self.authorized_keys
+        if public_keys == current:
+            return
+
+        self._authorized_path.write_text("\n".join(public_keys) + "\n")
+
+    def keygen(self, renew: bool = False, keysize: int = 4096) -> None:
```
marcoppenheimer, line 67: **suggestion:** It may be like over-cautious, but do we want a pw on the key?

marcoppenheimer (approved):

zmraul (commented):
Only comment would be to clarify what uses the Reconciliable class can have.

src/managers/ssh.py
imanenami: Unfortunately, `medusa` does not support password-protected keys yet: https://github.com/thelastpickle/cassandra-medusa/blob/f4ce57622030c593e9640f97d1fc73073899611e/docs/Configuration.md?plain=1#L141

(expected in PEM unencrypted format)

src/events/__init__.py
imanenami: I had bigger plans for this to add some type-hint values, but it required more unrelated refactoring so I aborted in the middle. I'll remove this since it's useless now :+1:

src/events/cassandra.py
imanenami: Good catch, it is not needed

zmraul (approved):

PR #61, opened 2026-04-09
[DPE-9748] feat: Initial backups implementation

imanenami:
This PR implements the backbone of backup/restore operations, backed with **manual testing** done using AWS S3:

- Install & configure `medusa` on units
- Adds `s3-credentials`, `gcs-credentials`, and `azure-storage-credentials` interfaces and the supporting charmlib
- Adds `create-backup`, `list-backups`, and `restore` actions
- Implements `BackupManager` which uses `medusa` to manage backup/restore ops
- Implements `BackupEvents` which handle backup/restore actions. `restore` action handler is dummy atm.


src/core/state.py
```diff
@@ -316,6 +322,87 @@ def resolved(self) -> ResolvedTLSContext:
...
+    # Generic
+
+    @property
+    def secret_key(self) -> str:
+        """S3 secret access key or GCS secret key JSON."""
+        raw = self.relation_data.get("secret-key", "")
+        if isinstance(raw, str):
+            return raw
+        else:
+            return json.dumps(raw)
```
marcoppenheimer, line 356: **question:** Is this because the libs handle the return-types differently?

src/core/workload.py
```diff
@@ -89,6 +89,16 @@ def jmx_exporter_config(self) -> pathops.PathProtocol:
...
 
+    @property
+    def medusa_config(self) -> pathops.PathProtocol:
+        """Path to Medusa config file."""
+        return self.config_dir / "medusa.ini"
+
+    @property
+    def storage_credentials(self) -> pathops.PathProtocol:
+        """Path to Medusa config file."""
+        return self.config_dir / "credentials"
```
marcoppenheimer: **big nitpick:** Does `credentials` here need to be hardcoded? It's a vague name if it's a shared conf dir

src/events/backup.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+        self.framework.observe(
+            getattr(self.charm.on, "list_backups_action"), self._on_list_backups_action
+        )
+        self.framework.observe(getattr(self.charm.on, "restore_action"), self._on_restore_action)
+
+    def _on_create_backup_action(self, event: ActionEvent) -> None:
+        """Handle the `create-backup` Juju action."""
+        if not self._run_before_checks(event):
+            return
```
marcoppenheimer, line 85: **question:** What's the behavior when we `juju run cassandra create-backup` on the whole app? Do we need a 'leader-only' check?

src/events/backup.py
```diff
@@ -0,0 +1,204 @@
...
+            return False
+
+        if not self._check_units():
+            logger.error(f"precheck failed: {BackupMessages.WORKLOAD_NOT_READY.value}")
+            event.fail(BackupMessages.WORKLOAD_NOT_READY.value)
+            return False
+
+        return True
+
+    def _check_units(self) -> bool:
```
marcoppenheimer, line 134: **suggestion:** Should we also have a 'is refreshing' check?

src/managers/backup.py
```diff
@@ -0,0 +1,198 @@
...
+    """Data container for backups."""
+
+    name: str
+    started: datetime.datetime
+    finished: datetime.datetime | None
+    state: Literal["finished", "incomplete"]
+    mode: BackupMode = "full"
+
+    def as_da096_dict(self, base_repo_url: str) -> dict[str, str | float | None]:
+        """Serialize as a DA-096 compatible dict."""
```
marcoppenheimer: **todo:** I know I'm being a bit pedantic with the names, but lets not require users to read a spec to understand what this is.
`as_dict` is fine, and if we REALLY want, we can model this dict somewhere (pydantic?)

src/events/backup.py
marcoppenheimer: **question:** Given that this may take some time for heavy nodes, is there anything we can do before the event exits in the `juju status` showing that 'hey, we might be here a while - doing a backup'?

src/managers/backup.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if "aws" in self.host:
+            return "s3"
+        else:
+            return "s3_compatible"
+
+    @property
+    def secure(self) -> bool:
+        """Is the storage using HTTPS?"""
+        return True if self.parsed_endpoint.scheme == "https" else False
```
marcoppenheimer: **big nitpick:** `return self.parsed_endpoint.scheme == "https"`

src/managers/backup.py
```diff
@@ -0,0 +1,198 @@
...
+        stdout = self.medusa_exec("list-backups", timeout=60)
+        ret = []
+        for line in stdout.split("\n"):
+            if not line:
+                break
+
+            # complete line sample:
+            # {backup_id} (started: {ISO DT str}, finished: {ISO DT str})
+            # incomplete line sample:
+            # {backup_id} (started: {ISO DT str}, finished: Incomplete [2 of 3 nodes finished])
```
marcoppenheimer, line 133: **praise:** Good comment, especially helpful for painful parsing

src/managers/backup.py
```diff
@@ -0,0 +1,198 @@
...
+            "nodetool_password": config.nodetool_password,
+            "storage_provider": config.storage_provider,
+            "storage_bucket": config.storage_bucket,
+            "storage_region": config.storage_region,
+            "storage_host": config.host,
+            "storage_secure": config.secure,
+        }
+        self._workload.cassandra_paths.medusa_config.write_text(template.render(data) + "\n")
+
+    def _render_s3_credentials(self, access_key: str, secret_key: str) -> None:
```
marcoppenheimer, line 178: **question:** Do we not need to render for GCP? Or does GCS also use `aws_access_key_id` as it's S3 compatible?

tests/integration/test_backup_microceph.py
```diff
@@ -0,0 +1,160 @@
...
+
+def test_list_backups_succeeds_and_is_empty(juju: jubilant.Juju, app_name: str):
+    task = juju.run(f"{app_name}/0", "list-backups")
+    assert task.results.get("result") == "[]"
+
+
+def test_create_backup(juju: jubilant.Juju, app_name: str):
+    juju.wait(all_active_idle)
+    task = juju.run(f"{app_name}/0", "create-backup", wait=1200)
+    assert task.results.get("backup-status") == "backup created"
```
marcoppenheimer, line 142: **todo:** Can we do more than just list-backups here, and 'GET' data from Ceph? Just checking existence of data should be fine

marcoppenheimer (approved):
Looks good, simpler than I expected but still with a lot of moving parts - nice work.

Left some non-blocking comments and questions that can be fairly easily addressed.

src/core/state.py
imanenami: Yes, the GCS secret key is a JSON object for instance

src/core/workload.py
imanenami: It's actually a file path, which is used in medusa config

src/events/backup.py
imanenami: afaik, and correct me if I'm wrong, juju actions can be run on units (not apps). The UX here is the same as postgres, the unit shows `(create-backup)` and remains in `executing` state until the backup process is finished. IMO, an additional status would be redundant, and complicates the handler logic.

Based on the original spec, backup/restore could be run on any unit, non-leader preferred.

src/events/backup.py
imanenami: This is embedded in the `is_operational` check :)

src/managers/backup.py
imanenami: Yeah maybe the docstring is enough 👍

src/managers/backup.py
imanenami: Based on the lib, GCS only needs the `secret_key` which itself is a JSON, and the logic is handled in https://github.com/canonical/cassandra-operator/pull/61/changes/BASE..52d37814a5d6cbcccde386a85d8d471ae24907cf#diff-52276baed1b3ec79e2933978e9fe3d075475bb8c58f189d1a538f188a1b9ed8dR157-R160

tests/integration/test_backup_microceph.py
imanenami: Sure, good idea. But the better test would be to restore from the backup, which also implicitly tests what you mentioned, and it's added in subsequent PRs. so if you don't mind, I don't touch the tests here in favor of better tests there :)

actions.yaml
```diff
@@ -54,4 +54,24 @@ resume-refresh:
...
+# Backup/Restore Actions
+create-backup:
+  description: Creates a backup to s3 storage.
+  params:
+    type:
+      type: string
+      description: The backup type, the default value is 'full'.
+        Full backup is a full copy of all data.
+        Differential backup is a copy only of changed data since the last full backup.
+        Incremental backup is a copy only of changed data since the last backup (any type).
```
zmraul: **nit**: Are these three going to be implemented in the short term? If not, probably best to just mention the full.

src/workload.py
```diff
@@ -151,3 +156,31 @@ def exec(
...
+        self.exec(["apt", "install", "-y", "python3-pip"])
+        self.exec(
+            [
+                "pip",
+                "install",
+                "-I",
+                "--break-system-packages",
+                "cassandra-medusa[S3]==0.27.1",
+            ]
+        )
```
zmraul, line 174: **todo**: this needs to be tracked on a repo issue, so it's not hidden away in here.

src/events/backup.py
```diff
@@ -0,0 +1,200 @@
...
+        return True
+
+    @property
+    def active_context(self) -> StorageClientContext | None:
+        """Return the active context based on storage relations state."""
+        for ctx in [self.s3_context, self.gcs_context, self.azure_storage_context]:
+            if ctx.ready:
+                return ctx
+
+        return None
```
zmraul, line 162: **question**: this is not excluding several relations being present at once no? Is the current approach to allow them, and default to s3 as main one?

src/events/backup.py
```diff
@@ -0,0 +1,200 @@
...
+            event.fail(BackupMessages.OP_FAILED.value)
+
+    def _on_list_backups_action(self, event: ActionEvent) -> None:
+        """Handle the `list-backups` Juju action."""
+        if not self._run_before_checks(event):
+            return
+
+        try:
+            backups = self.backup_manager.list_backups()
+            repo = f"{self.s3_context.endpoint}/{self.s3_context.bucket}"
```
zmraul: **todo**: Listing only works on s3. Should this grab from `active_context` instead?

src/managers/backup.py
```diff
@@ -0,0 +1,198 @@
...
+        """Is the storage using HTTPS?"""
+        return self.parsed_endpoint.scheme == "https"
+
+
+class BackupManager:
+    """Manager of medusa-driven backup/restores."""
+
+    def __init__(self, workload: WorkloadBase, s3_endpoint: str | None = None):
+        self._workload = workload
+        self._endpoint = s3_endpoint
```
zmraul: **todo**: `s3_endpoint` is not used

src/events/backup.py
```diff
@@ -0,0 +1,200 @@
...
+            for path in [
+                self.workload.cassandra_paths.medusa_config,
+                self.workload.cassandra_paths.storage_credentials,
+            ]:
+                if path.exists():
+                    path.unlink()
+            return
+
+        if not self.active_context:
+            return
```
zmraul, line 190: **nit**: already part of `self.ready` check, can be removed

src/events/backup.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def __init__(
+        self,
+        charm: TypedCharmBase[CharmConfig],
+        state: ApplicationState,
+        workload: WorkloadBase,
+        node_manager: NodeManager,
+        ssh_manager: SSHManager,
+    ):
+        super().__init__(charm, key="provider_events")
```
zmraul: **todo**: `backup_events` or something like that. I'm actually surprised that duplicating this with the `provider` file doesn't seem to affect the charm at all given the tests were passing. 🤔

src/events/backup.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env python3
+# Copyright 2025 Canonical Ltd.
```
zmraul:
```suggestion
# Copyright 2026 Canonical Ltd.
```

src/events/backup.py
```diff
@@ -0,0 +1,200 @@
...
+# See LICENSE file for licensing details.
+
+"""Handler for main Cassandra charm events."""
```
zmraul:
```suggestion
"""Handler for backup Cassandra charm events."""
```

src/events/backup.py
```diff
@@ -0,0 +1,200 @@
...
+logger = logging.getLogger(__name__)
+
+
+class BackupMessages(str, Enum):
+    """Enum for backup/restore messages."""
+
+    NOT_READY = (
+        "Can not initiate backup/restore operation. "
+        "Check if s3-integrator is in active|idle state, "
+        "and if the charm is integrated properly with s3-integrator."
```
zmraul:
```suggestion
        "Check if storage relation is in active|idle state, "
        "and if the charm is integrated properly with the storage integrator."
```

zmraul (changes requested):
Thanks! Left some comments. There are some minor issues, but looks good overall.

~~The feature is missing unit tests, you could check [opensearch ones](https://github.com/canonical/opensearch-operator/blob/main/tests/unit/lib/test_backups.py). But the PR is big enough as it is, so probably best to do as follow up.~~ Already addressed

src/events/backup.py
imanenami: Based on discussion with Mykola, we want to support multiple storage relations in the future. so yes.

actions.yaml
imanenami: Full for now :) will fix it :+1:

src/events/backup.py
imanenami: The repo is imo a useless field in list-backups output, but will fix it :)

src/events/backup.py
imanenami: Yes seen that before, the key conflict doesn't have any effect, but will fix it :+1:

src/events/backup.py
imanenami: The linter is not that smart unfortunately :( so will keep it

src/events/backup.py
imanenami: Applied https://github.com/canonical/cassandra-operator/pull/61/commits/ec2426a1238ff9b15b207ef94492c2ad5bc17513

src/events/backup.py
imanenami: Applied https://github.com/canonical/cassandra-operator/pull/61/commits/ec2426a1238ff9b15b207ef94492c2ad5bc17513

src/managers/backup.py
imanenami: Fixed here: https://github.com/canonical/cassandra-operator/pull/61/commits/ec2426a1238ff9b15b207ef94492c2ad5bc17513****

actions.yaml
imanenami: Fixed here: https://github.com/canonical/cassandra-operator/pull/61/commits/ec2426a1238ff9b15b207ef94492c2ad5bc17513

zmraul (approved):

src/core/workload.py
marcoppenheimer: I mean if we `cd` to `$CONF` and we see `$CONF/credentials` file, it's not clear which creds they are no?

imanenami: Ah think got your point. In fact `credentials` is a standard name in AWS cli context, but not for Azure/GCS. Do you think simple renaming to `storage-credentials` or such addresses your concern? I'm hesitant to make different files for different storage providers at this stage, specially since the purpose is the same...

src/workload.py
imanenami: Thanks, opened an issue to track this:
https://github.com/canonical/cassandra-operator/issues/70

src/core/workload.py
marcoppenheimer: storage-credentials sounds good

marcoppenheimer (approved):

PR #65, opened 2026-04-14
[DPE-9777] feat: add restore functionality

imanenami:
This PR adds restore functionality and associated integration tests.


src/events/backup.py
```diff
@@ -46,11 +54,43 @@ class BackupMessages(str, Enum):
...
+        backup_name: str,
+        app: Application | None = None,
+        unit: Unit | None = None,
+    ):
+        super().__init__(handle, relation, app=app, unit=unit)
+        self.backup_name = backup_name
+
+    def snapshot(self) -> dict:
+        """Return a snapshot of the event."""
+        return super().snapshot() | {"backup-name": self.backup_name}
```
marcoppenheimer, line 77: **question:** It's been a while since I got upset with relation-data, why do we need to add this extra thing to the snapshot?

src/events/backup.py
```diff
@@ -119,7 +160,38 @@ def _on_restore_action(self, event: ActionEvent) -> None:
...
+
+        if not (backup_name := event.params.get("backup-id")):
+            event.fail(BackupMessages.INVALID_BACKUP_ID.value)
+            return
+
+        self.state.unit.restoring = True
+        self.state.unit.restore_backup_name = backup_name
+        self.state.unit.workload_state = UnitWorkloadState.RESTORING
+        event.set_results({"result": "restore started."})
+
```
marcoppenheimer, line 175: **question:** Are we actually restoring here? It reads like we're writing relation-data, is it triggered indirectly by a relation-changed?

src/events/cassandra.py
```diff
@@ -626,6 +639,23 @@ def _repair_auth(self, event: EventBase) -> bool:
                 self.state.unit.auth_repaired = False
         return True
 
+    def _update_restoring_state(self) -> None:
+        """Update the restoring state of the unit & cluster."""
+        if self.state.restoring:
+            self.state.unit.workload_state = UnitWorkloadState.RESTORING
+            if self.state.unit.restore_backup_name:
```
marcoppenheimer, line 646: **nitpick:** We can probably put both these conditions on the same line

src/events/cassandra.py
```diff
@@ -361,6 +364,11 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
             )
             event.defer()
             return
+
+        self._update_restoring_state()
+        if self.state.restoring:
+            return
```
marcoppenheimer, line 370: **note:** Ah, it is

src/events/cassandra.py
```diff
@@ -408,6 +416,8 @@ def _on_update_status(self, _: UpdateStatusEvent) -> None:
             logger.error("Refresh manager is not ready")
             return
 
+        self._update_restoring_state()
+
```
marcoppenheimer, line 420: **question:** So this feels strange to me.

If I understand correctly, we handle the `restore` action, state to the user 'yeah, we're restoring now', and then on basically any other event, we attempt a restore.

If that's correct, why don't we just emit the restore during the `restore` action event? It would effectively block the charm, but is that not desired?

src/managers/backup.py
```diff
@@ -111,6 +112,13 @@ def medusa_exec(self, *args: str, timeout: int = 3600) -> str:
         )
         return stdout
 
+    def medusa_running(self) -> bool:
+        """Is a medusa process running?"""
+        raw, _ = self._workload.exec(["ps", "-eaf"])
```
marcoppenheimer, line 117: **nitpick:** We could just filter for `medusa` in the exec

src/managers/backup.py
```diff
@@ -148,6 +156,15 @@ def list_backups(self) -> list[BackupInfo]:
 
         return ret
 
+    def restore(self, backup_name: str) -> None:
+        """Restore a backup."""
+        default_args = ["--keep-auth", "-y", "--verify"]
+        if self.medusa_running():
+            logger.info("Medusa process in running. Waiting...")
```
marcoppenheimer, line 163: **todo:** `is running`
**question:** I'm assuming that Medusa 'running' means that a back-up is in progress? If so, we can probably change the terminology here and elsewhere to just say 'is backup running'

src/events/backup.py
```diff
@@ -119,7 +160,38 @@ def _on_restore_action(self, event: ActionEvent) -> None:
...
             return
 
-        logger.info(f"restore {repr(event)}")
+        if self.state.restoring:
+            event.fail(BackupMessages.RESTORE_IN_PROGRESS.value)
+            return
+
+        if not (backup_name := event.params.get("backup-id")):
+            event.fail(BackupMessages.INVALID_BACKUP_ID.value)
+            return
```
marcoppenheimer, line 169: **todo:** I think we depend on these conditionals a lot, we should add basic unit-tests for them.

src/events/backup.py
```diff
@@ -119,7 +160,38 @@ def _on_restore_action(self, event: ActionEvent) -> None:
...
+
+        try:
+            logger.info(f"Initiating restore on {self.state.unit.unit_name}")
+            self.backup_manager.restore(backup_name=event.backup_name)
+            logger.info("Restore finished successfully.")
+        except ExecError as e:
+            logger.error(f"Restore process failed: {e.stdout} {e.stderr}")
+        finally:
+            self.state.unit.restoring = False
+            self.state.unit.restore_backup_name = ""
```
marcoppenheimer, line 194: **todo:** Same here, unit-tests for the conditions

src/events/cassandra.py
marcoppenheimer: **todo:** Unit-test

src/managers/backup.py
```diff
@@ -111,6 +112,13 @@ def medusa_exec(self, *args: str, timeout: int = 3600) -> str:
         )
         return stdout
 
+    def medusa_running(self) -> bool:
```
marcoppenheimer, line 115: **suggestion:** Less important here as it's mostly exactly what it looks like, but might be worth just adding a basic test with some mock `ps` outputs

marcoppenheimer (changes requested):

src/events/backup.py
imanenami: The whole backup/restore implementation requires improving unit test coverage :+1: I'll add them

src/events/cassandra.py
imanenami: The issue here is that the restore action could run on any unit, and we want the whole app/units enter into a maintenance status so that no other juju actions are run during the restore op. If we block on the action handler, other units can't get notified of the impending restore. Of course, the workload (Cassandra database) operates normally, and we only fence Juju actions.

src/events/backup.py
imanenami: The `super` method does not take care of the additional `backup_name` attribute added, so we need to override that behavior

src/events/backup.py
imanenami: Please refer to my answer here: https://github.com/canonical/cassandra-operator/pull/65#discussion_r3090947227

src/events/cassandra.py
marcoppenheimer: Ack, makes sense, no problems.

imanenami:
I added unit tests to this PR: https://github.com/canonical/cassandra-operator/pull/69 alongside the cloud tests

I tried to focus more on real unit testing rather than the so-called scenario tests unless I had no choice :)

The reason of my maximum scenario test avoidance is due to the nature of `ops-scenario` lib and the couple of instances in the past where its behavior has diverged from real ops behavior. I think we should adopt a pure unit testing approach in our current & future tests wherever possible

src/managers/backup.py
imanenami: Possibly resolved by https://github.com/canonical/cassandra-operator/pull/65#issuecomment-4266647986

src/events/backup.py
imanenami: Possibly resolved by https://github.com/canonical/cassandra-operator/pull/65#issuecomment-4266647986

src/events/backup.py
imanenami: Possibly resolved by https://github.com/canonical/cassandra-operator/pull/65#issuecomment-4266647986

src/events/cassandra.py
imanenami: Since we want to set a workload state on all units, and only run the restore command on one unit, I think it's not possible.

src/managers/backup.py
imanenami: Fixed here: https://github.com/canonical/cassandra-operator/pull/65/commits/8ea18377a2fb9d2f365e1ad050638982e68cad0f

marcoppenheimer:
> couple of instances in the past where its behavior has diverged from real ops behavior

That's surprising, can you share some examples that you saw?
Personally I agree it's more valuable to validate the event flow with int-tests, but it can definitely be helpful to verify events exit/trigger things when relation-data conditions aren't met, so I'd be really interested to see reason to not use it.

marcoppenheimer (approved):

src/events/backup.py
```diff
@@ -116,7 +157,38 @@ def _on_restore_action(self, event: ActionEvent) -> None:
...
+        if self.backup_manager.medusa_running():
+            event.defer()
+            return
+
+        try:
+            logger.info(f"Initiating restore on {self.state.unit.unit_name}")
+            self.backup_manager.restore(backup_name=event.backup_name)
+            logger.info("Restore finished successfully.")
+        except ExecError as e:
+            logger.error(f"Restore process failed: {e.stdout} {e.stderr}")
```
zmraul, line 188: **question**: Is this the current UX around restore? If something goes wrong, the only place it shows is on the logs. I'm guessing yes, since the action is async, but worth checking.

src/events/backup.py
```diff
@@ -116,7 +157,38 @@ def _on_restore_action(self, event: ActionEvent) -> None:
...
+            event.fail(BackupMessages.INVALID_BACKUP_ID.value)
+            return
+
+        self.state.unit.restoring = True
+        self.state.unit.restore_backup_name = backup_name
+        self.state.unit.workload_state = UnitWorkloadState.RESTORING
+        event.set_results({"result": "restore started."})
+
+    def _on_restore_event(self, event: RestoreEvent) -> None:
+        """Handle custom `restore` event."""
```
zmraul, line 174: Probably good to have some health checks here, to short circuit the whole flow?

zmraul (approved):
Neat

src/events/backup.py
imanenami: The existing `if not self.state.unit.restoring or not self.state.restoring` check is the quickest I could think of, it only depends on relation data :)

PR #68, opened 2026-04-20
[DPE-9794] fix: premature bootstrap issues

### LINKED ISSUE #62: Possibe race condition on multi-unit deploy bootstrap process
There's a possible deadlock scenario, which could be because of a race condition, when attempting to deploy a multi-unit Cassandra app. It does happen sometimes and I've seen it both in manual and CI tests (sample failing run included below).

## Steps to reproduce
```
charmcraft pack
juju deploy ./cassandra_ubuntu@24.04-amd64.charm --config profile=testing -n 3
```

## Expected behavior
Units should go into `active|idle` state after some time.

## Actual behavior
Occasionally, the units are stuck in maintenance mode forever, with status messages such as `repairing system_auth keyspace` or `installing Cassandra`. The latter message is misleading, because the Cassandra snap is actually installed and present on the unit reporting it.

## Versions

Operating system: Ubuntu 24.04 LTS
Juju CLI: 3.6.21 
Juju agent: 3.6.21 
Charm revision: 

## Log output

Sample failing CI run with all the required info:
https://github.com/canonical/cassandra-operator/actions/runs/24205529405/job/70662440763

## Additional context

I suspect the issue is in `common.lock_manager.LockManager` implementation, but take this with a grain of salt.

imanenami:
## Description

- Fixes https://github.com/canonical/cassandra-operator/issues/62
- Fixes another deadlock issue I discovered while testing, in complex TLS scenario when adding new units

## Findings

While the implementation of the core and managers logic is exemplary, the current implementation of the event handlers is an ***Event Maze***, where slight changes in firing order of event/deferred events might lead to unexpected state. The long term solution imo is to adopt a ***Reconciler*** pattern

### DIFF
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -512,7 +512,7 @@ class CassandraEvents(Object):
     @property
     def _are_seeds_reachable(self) -> bool:
         return self.state.unit.is_seed or any(
-            unit.is_operational and self.database_manager.check(hosts=[unit.ip])
+            unit.is_operational and self.workload.ping(hosts=[unit.ip])
             for unit in self.state.other_seed_units
         )
 
--- a/src/events/cassandra.py
+++ b/src/events/cassandra.py
@@ -39,6 +39,7 @@ from core.state import (
     PEER_RELATION,
     ApplicationState,
     AuthRepairState,
+    ClusterState,
     UnitWorkloadState,
 )
 from core.statuses import Status
@@ -234,13 +235,13 @@ class CassandraEvents(Object):
 
     def _start_subordinate(self, event: StartEvent) -> None:
         if not self.state.cluster.is_active and not self.state.unit.is_seed:
-            self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
+            # self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
             logger.debug("Deferring subordinate on_start due to cluster not being active yet")
             event.defer()
             return
 
         if not self._are_seeds_reachable:
-            self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
+            # self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
             logger.debug("Deferring subordinate on_start due to seeds not being ready")
             event.defer()
             return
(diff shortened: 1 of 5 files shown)


src/events/cassandra.py
```diff
@@ -215,13 +216,13 @@ def _start_leader_setup_auth(self) -> None:
 
     def _start_subordinate(self, event: StartEvent) -> None:
         if not self.state.cluster.is_active and not self.state.unit.is_seed:
-            self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
+            # self.state.unit.workload_state = UnitWorkloadState.WAITING_FOR_START
```
imanenami, line 238: Setting the state here conflicts with the expected state on bootstrap, leading to deadlock on non-leader units. If setting this state is important (which is not imo), it should be set on collect status.

src/events/cassandra.py
```diff
@@ -490,7 +493,7 @@ def _on_collect_app_status(self, event: CollectStatusEvent) -> None:
     @property
     def _are_seeds_reachable(self) -> bool:
         return self.state.unit.is_seed or any(
-            unit.is_operational and self.database_manager.check(hosts=[unit.ip])
+            unit.is_operational and self.workload.ping(hosts=[unit.ip])
```
imanenami, line 515: When new unit is being added to a complex TLS deployment (by complex, I mean using provided peer & client certs), the unit is not able to pass the `database_manager.check` due to TLS not being ready, leading to deadlock in the bootstrap. In this stage, the seeds passing socket check is more than enough.

src/events/cassandra.py
```diff
@@ -582,3 +585,23 @@ def _repair_auth(self, event: EventBase) -> bool:
             else:
                 self.state.unit.auth_repaired = False
         return True
+
+    def _handle_premature_leader_bootstrap(self) -> None:
```
zmraul, line 608: **question**: why call this on `update_status`. Isn't it better to have it somewhere along the start sequence, or near a reconcile pattern? If the deadlock happens we still have to wait to update_status which might take long time to arrive.

imanenami: I agree with your point that this should be called on a reconcile method, but the current Cassandra charm implementation lacks one. 

Based on my evaluation of the codebase, adoption of `Reconciler` pattern requires dedicated effort so as not to break things, much of which focused at fixing the `ops-scenario` house of cards tests.

As mentioned in the PR description, I prefer & suggest to keep this bugfix PR simple, and add a backlog issue for adopting Reconciler pattern...

src/events/cassandra.py
marcoppenheimer: We can probably drop it. There's likely a correct term for it, but I'd rather track states that describe current state, rather than incoming states

i.e `initialised` rather than `waiting to start`

marcoppenheimer (approved):

zmraul (approved):

